PR #1, opened 2019-12-23
Add a starting point for a contributor's document

jffng:
Starting with a minimal set of guidelines for now, but it may be nice to include some suggestions as to what kinds of contributions could be interesting.


kjellr:
🚢 Let's work on getting a thorough README in here next. Thanks @jffng!

PR #5, opened 2020-01-15
Create the TwentyTwenty Demo Site using Block Templates

jffng:
**Description**

[_WIP_] This PR is an attempt to create https://2020.wordpress.net/ using block templates. The purpose is to demonstrate how block templates work and identify any hiccups encountered developing a theme + site this way.

**How To Test**

2. Install Gutenberg and enable the Full Site Editing experiment
3. Upload the demo content (may replace this with starter content later) 
4. Clone, install, and activate this theme

**Known Issues**
- Padding / spacing in header and footer 
- Home page design is missing the background white element on the featured image
- Mobile menu
- Search bar

**Screenshots**

_Home page_
Screen Shot 2020-01-15 at 11 16 31 AM

_Exhibitions page_
Screen Shot 2020-01-15 at 11 17 17 AM

_Footer_
Screen Shot 2020-01-15 at 11 17 36 AM


twentytwenty-blocks/style.css
```diff
@@ -0,0 +1,76 @@
+/*
+Theme Name: Twenty Twenty
```
kjellr:
```suggestion
Theme Name: Twenty Twenty Blocks
```

^ Updated to differentiate from the existing version of Twenty Twenty.

twentytwenty-blocks/style.css
```diff
@@ -0,0 +1,76 @@
...
+Author: the WordPress team
+Author URI: https://wordpress.org/
+Description: Twenty Twenty, but only using block templates.
+Requires at least: WordPress 4.9.6
+Version: 0.03
+License: GNU General Public License v2 or later
+License URI: LICENSE
+Text Domain: twentytwenty
```
kjellr:
```suggestion
Text Domain: twentytwenty-blocks
```
^ Update to match the theme's folder name.

twentytwenty-blocks/readme.txt
```diff
@@ -0,0 +1,98 @@
+=== Twenty Twenty ===
+Contributors: the WordPress team
+Tested up to: 5.3
+Stable tag: 1.1
+License: GPLv2 or later
+License URI: http://www.gnu.org/licenses/gpl-2.0.html
+
+Default theme for 2020.
```
kjellr: This should be updated a bit so that it's clear it's an experiment and not the actual 2020 theme. Can also include instructions for installing it so that the starter content kicks in. You can check out what I did for this in #4:

https://github.com/WordPress/theme-experiments/blob/16f681b6beaa27efef4aa028e391b424f13339ab/twentynineteen-blocks/readme.txt

twentytwenty-blocks/block-templates/404.html
```diff
@@ -0,0 +1,11 @@
+<header class="site-header">
```
kjellr: Is this file required? It appears to be identical to `index.html`.

jffng:
@kjellr I made two commits that attempt to simplify the example by moving the blocks that were previously in block templates, directly into the starter content. 

This has caused some unexpected issues that I could use a second pair of eyes on to debug, namely: 

- When I activate the theme, only the **Home**, **Contact**, and **Blog** pages are created with the correct content. 
- **About** is created but missing the starter content I've specified. Maybe this has to do with the theme supplying a `page-about.html`.
- **Exhibitions** and **From Signac to Matisse** are not created at all. 

I've tested this locally and on jurassic ninja.

kjellr:
@jffng: I pushed a bunch of updates this morning. These sort of sidestep [the issues you'd been having around the starter content](https://github.com/WordPress/theme-experiments/pull/5#issuecomment-575836504) by (mostly) falling back on the starter content that's baked into the theme by default, rather than the more complicated starter content in the theme demo site. This is a little bit of a shift from the initial attempt that this PR made, but I think it is a decent start for the demo, as it simplifies things greatly. 

I also updated the `readme.txt`, and fixed up some styles around the page header. These aren't exactly like the original theme, but they're pretty close. The main hiccup is the treatment around the featured image. We can't easily add a white background behind it on the homepage (since we're not _actually_ displaying the real featured image yet), and we can't replicate the full-screen featured image that lives behind the post title on subpages with blocks yet either. 

Here are a few screenshots of a fresh install, with the starter content active:

Screen Shot 2020-01-20 at 09 45 38
 
Screen Shot 2020-01-20 at 09 46 55

Screen Shot 2020-01-20 at 09 47 17

---

Anyway, let me know what you think. If this is generally working, I think we can probably merge this in and iterate so other folks can get testing more easily.

twentytwenty-blocks/functions.php
```diff
@@ -0,0 +1,309 @@
...
+			// Specify the core-defined pages to create and add custom thumbnails to some of them.
+			'posts' => array(
+				'front' => array(
+					'post_type'    => 'page',
+					'post_title'   => __( 'The New UMoMA Opens its Doors', 'twentytwenty-blocks' ),
+					// Use the above featured image with the predefined about page.
+					'thumbnail'    => '{{image-opening}}',
+					'post_content' => join(
+						'',
+						array(
```
kjellr: I'm a little torn about how to handle this starter content still: 

For our purposes now, I do think it would help code readability to move this into template part, but I'm not sure it actually makes sense to at this point — since this is just standard page content, users will want to edit this inside the standard page editor, not in the block templates area of the editor.

kjellr: I think what we're really looking to use is "Page Templates", as per https://github.com/WordPress/gutenberg/issues/18055. That functionality doesn't appear to be ready for themes too hook into yet, so sticking with the starter content is probably the best option we have at the moment. I've left some feedback in that ticket here: https://github.com/WordPress/gutenberg/issues/18055#issuecomment-576336048.

jffng: While it does improve readability at this moment, I agree that putting starter content in a block template is not the right approach.

jffng:
It works like a charm.

Much simpler and easier to understand the concept of a block based theme this way, instead of trying to recreate the demo site entirely, thank you!  🎉

PR #11, opened 2020-01-22
Update README.md

kjellr:
This PR adds: 

- Installation instructions
- A warning about use on a production site
- A questions & improvements section.

It also trims down the resources list to focus just on official sources for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,9 +3,11 @@
 ⚠️ These themes are not intended for use on a production site.
 
 ## About
+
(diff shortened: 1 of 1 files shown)


kjellr:
> Based on the php version issue yesterday, do you think it's worthwhile to also helpful to include up front the environment details that these themes are being developed on, e.g.

Sure — since WordPress supports PHP 7.2, we'll want to support that for sure. But I did add some recommendations in the doc.

PR #16, opened 2020-01-31
[Gutenberg Starter Theme] Try pulling starter content in from an external file

kjellr:
This PR tries out an idea that came up during a discussion with @epiqueras yesterday — that starter content could theoretically be pulled in from a standalone "block template"-style file, rather than being declared as a complicated array in PHP, as is done in themes like [this repo's version of Twenty Twenty](https://github.com/WordPress/theme-experiments/blob/bc79700be1ea4447ed94fed01610b570f3ac3900/twentytwenty-blocks/functions.php#L108-L220). 

This PR creates a new `block-starter-content` folder for the Gutenberg Starter Theme (Blocks). It contains a `post-demo.html` file, full of the content from Gutenberg's demo post. This content gets pulled into the homepage via `file_get_contents()` for now. I'm not sure if that's the right function, or if it needs to be escaped and/or limited in some way... I could see this being problematic if the file size were huge. But it appears to work for this example at least. I'd love some more experienced PHP 👀 on that method. 

To test: 

1. Start a brand new WordPress site.
2. Clone this repository or upload a zip of the `Gutenberg Starter Theme (Blocks)` into your site’s themes directory.
3. Install the Gutenberg plugin.
4. Enable the "Full Site Editing" experiment in the Gutenberg > Experiments panel.
5. Activate the new theme from within Appearance > Themes.
6. Visit the Customizer to automatically load the demo content. 

The homepage should now contain the demo content from `block-starter-content/post-demo.html`.

---

**Screenshot**

Screen Shot 2020-01-31 at 4 11 30 PM


jffng:
@kjellr shall we close this? Seems like content like this, if at all, could ship with a pattern.

kjellr:
Yeah, sounds good to me. 👍

PR #17, opened 2020-02-11
Try/global styles support

nosolosw:
Work In Progress - see an alternative at https://github.com/WordPress/theme-experiments/pull/22

This is an experiment to try global styles on the Parisienne theme.

- Check out this pull request locally.
- Follow instructions at https://github.com/WordPress/gutenberg/pull/20047

### DIFF
--- a/parisienne/experimental-theme.json
+++ b/parisienne/experimental-theme.json
@@ -1,0 +1,14 @@
+{
+	"color": {
+		"foreground": "#141414",
+		"background": "#F3EBE8",
+		"button-text": "white",
+		"button": "#817359",
+		"border": "#DECABF",
+		"border-pre": "#EEE",
+		"link": "#59BACC",
+		"nav-text-hover": "#AAA",
+		"nav-text": "black",
+		"text": "#666"
+	}
+}
--- a/parisienne/experimental-theme.json
+++ b/parisienne/experimental-theme.json
@@ -4,6 +4,7 @@
 		"background": "#F3EBE8",
 		"button-text": "white",
 		"button": "#817359",
+		"primary": "#817359",
 		"border": "#DECABF",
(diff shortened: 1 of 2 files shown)


parisienne/experimental-theme.json
```diff
@@ -0,0 +1,15 @@
+{
+	"color": {
+		"foreground": "#141414",
+		"background": "#F3EBE8",
+		"button-text": "white",
+		"button": "#817359",
```
youknowriad: Why do we have a "button" and "button-text" variables, these seem too specific to me. Why . not just use "primary" for buttons and "light-text" or something for the button text (same for nav).

I think we should try to limit the number of variables as much as possible and compute the others.

nosolosw: Those are good thoughts that I also agree with!

The first step was just extracting what Parisienne had and see how that looked. The next iteration for this PR should consolidate them into something general (for what we need to come up with a set of variables and its derivatives).

nosolosw:
Note that this is not ready for review! At the moment it's sort of a playground to demo Global Styles with `edit-site`. If anyone would add the "In Progress" label I'd be grateful (I don't have permissions to do it myself).

parisienne/experimental-theme.json
kjellr: I do agree that these seem overly specific for now. Feel free to just assign primary/background colors here. 

> I think we should try to limit the number of variables as much as possible and compute the others.

I agree that this makes sense for the defaults, but I do think it'll be important to make sure theme authors (and users) can override these computed values with custom ones on a theme + global styles level.

nosolosw:
Closing this in favor of https://github.com/WordPress/theme-experiments/pull/22

PR #20, opened 2020-02-17
Add a home page template using the latest posts block to 2019

youknowriad:
This PR tries to add a home page showing the list of posts to the 2019 theme using the recent posts block.

### DIFF
--- a/twentynineteen-blocks/style.css
+++ b/twentynineteen-blocks/style.css
@@ -113,11 +113,17 @@ Tags: one-column, flexible-header, accessibility-ready, custom-colors, custom-me
 }
 
 /* Latest posts block */
+
+.wp-block-latest-posts {
+	padding-left: 0;
+}
+
 .wp-block-latest-posts li > a {
 	font-size: 2.25em;
 	font-weight: 700;
 	font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
 	color: #111;
+	line-height: 1.2;
 }
 
 .wp-block-latest-posts.wp-block-latest-posts__list li > a::before {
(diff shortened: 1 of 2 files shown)


twentynineteen-blocks/block-templates/index.html
```diff
@@ -3,7 +3,9 @@
 <!-- /wp:group -->
 
 <!-- wp:group {"align":"full","className":"site-content"} -->
-<div class="wp-block-group alignfull site-content"><div class="wp-block-group__inner-container"><!-- wp:post-title /--><!-- wp:post-content /--></div></div>
+<div class="wp-block-group alignfull site-content"><div class="wp-block-group__inner-container">
+    <!-- wp:latest-posts {"postsToShow":100,"displayPostContent":true,"displayPostDate":true} /-->
```
youknowriad: I'm showing 100 posts now because pagination is not supported yet.

youknowriad: I'm also showing the excerpt, post date and title. Ideally, this would use a "Query" block where you could decide to show post comment count... (but the query block is not implemented yet).

twentynineteen-blocks/style.css
```diff
@@ -99,6 +99,14 @@ Tags: one-column, flexible-header, accessibility-ready, custom-colors, custom-me
 
 /* Post title positioning */
 
+@media only screen and (min-width: 768px) {
+	.wp-site-blocks .site-content > .wp-block-group__inner-container > *:not(.align-full):not(.align-wide) {
```
youknowriad: Not sure that is correct but I wanted to show all blocks in `site-content` as centered by default unless they have (align full/wide)

twentynineteen-blocks/style.css
```diff
@@ -112,6 +120,27 @@ Tags: one-column, flexible-header, accessibility-ready, custom-colors, custom-me
   }
 }
 
+/* Latest posts block */
+.wp-block-latest-posts li > a {
+	font-size: 2.25em;
+	font-weight: 700;
+	font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
+	color: #111;
+}
```
youknowriad: I'm trying to match the h2 style here

twentynineteen-blocks/block-templates/index.html
mtias: Definitely should be a Query since we need to read from the query context. It's also why I don't think adding pagination support to latest posts is a great idea: https://github.com/WordPress/gutenberg/pull/19969#issuecomment-586613632

twentynineteen-blocks/style.css
```diff
@@ -112,6 +120,27 @@ Tags: one-column, flexible-header, accessibility-ready, custom-colors, custom-me
   }
 }
 
+/* Latest posts block */
+.wp-block-latest-posts li > a {
```
mtias: This is a great example of the things we should be absorbing in the block style work. cc @ItsJonQ @nosolosw.

twentynineteen-blocks/style.css
kjellr: Yeah, this is right, but it makes the h1 rule below it unnecessary. I'll combine them and push a new update. 👍

twentynineteen-blocks/style.css
kjellr: 👍 Looks good. Just needed the line height, which I've added in 
f4da15b

twentynineteen-blocks/style.css
nosolosw: One potential answer for this would be this proposal https://github.com/WordPress/gutenberg/pull/20290

PR #21, opened 2020-02-17
Add Ambitious theme to Theme Experiments

Netzberufler:
I spent some time last week to create a block-based theme and felt it might be worth sharing it.

Things explored:
- Typical block templates like 404.html, archive.html, home.html, page.html and single.html
- More complex structure of block template parts using subfolders
- Extra home page template with blocks using the front-page.html
- Tried out the navigation block and added a basic mobile dropdown menu

Repo: https://github.com/Netzberufler/theme-experiments/tree/ambitious/ambitious


ambitious/block-template-parts/front-page/front-page-intro.html
```diff
@@ -0,0 +1,33 @@
...
+			<!-- wp:button {"textColor":"white","borderRadius":0,"className":"is-style-outline"} -->
+			<div class="wp-block-button is-style-outline"><a class="wp-block-button__link has-text-color has-white-color no-border-radius" href="https://github.com/Netzberufler/ambitious">Github Repository</a></div>
+			<!-- /wp:button -->
+
+		</div>
+		<!-- /wp:buttons -->
+
+	</div>
+</div>
+<!-- /wp:group -->
```
jffng: I ran into this same use case of wanting to include content in block templates. 

Since it's not re-used anywhere, I think this belongs in starter content rather than as a block template part. 

Some related context in the Gutenberg repo: https://github.com/WordPress/gutenberg/issues/18055

cc @kjellr

ambitious/block-template-parts/content/content-home.html
```diff
@@ -0,0 +1,11 @@
...
+<div class="wp-block-group type-post">
+	<div class="wp-block-group__inner-container">
+
+		<!-- wp:template-part {"slug":"entry/entry-title","theme":"ambitious"} /-->
+		<!-- wp:template-part {"slug":"entry/entry-meta","theme":"ambitious"} /-->
+		<!-- wp:post-excerpt /-->
+
+	</div>
+</div>
+<!-- /wp:group -->
```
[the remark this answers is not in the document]
Netzberufler: The content-home.html template is loaded from home.html and archive.html, so not used just once. But I should have named it content.html, got the name wrong. 

I have to admit I did not re-think the template structure much, I basically replicated the former PHP template files. The content-page.html is used from index.html and page.html. I guess the other extra templates like content-404.html and content-single.html are not necessary, although it is nice to have the same structure for all block templates.

ambitious/block-template-parts/front-page/front-page-intro.html
Netzberufler: Ah okay, good to know. I saw the about.html and contact.html files in the Parisienne theme and thought theme authors can now directly provide block templates for starter content.

I think this actually would be a really great feature. So much easier to provide both starter content and a predefined block layout in just one template file, instead of having two separate places. I will check out the issue on the Gutenberg project about that, haven't seen that one yet.

Netzberufler:
> 
> Nice contribution, @Netzberufler!
> 
> I left a couple comments, but in general I appreciate how you're pushing the use of block templates and template parts to cover more aspects of the traditional theme template hierarchy.
> 
> What were the challenges or desires you encountered in the process of converting your design to an entirely block-based theme?

@jffng Thanks so much for the quick response, I really appreciate it.

I think block templates already work great on the frontend and I had lots of fun building the theme. I can totally see myself using blocks all the way instead of PHP template tags and functions. It was also easier to switch to block templates than I thought beforehand. After playing around a bit with creating block layouts directly in the Editor I quickly jumped right into VS Code and wrote pure HTML. 

Here are some issues I ran into:

The Navigation block was difficult to style, since it required to override the default CSS. I think it would be great if the Core styling would be implemented as a block style, so theme authors can start with an unstyled HTML list and provide their own additional block styles, e.g. for different menus in the header and footer.

Putting the block template parts into subfolders worked fine in the frontend, but the Editor on `Gutenberg > Site Editor (beta) ` did not work at all with it.

It would be nice if theme authors had some Container Block which allows to add semantic markup like `<header>` and `<footer>`. I also felt that the Group block generates a lot of HTML markup, especially if you just want to wrap smaller elements like the post date.

I can see the need to be able to hide blocks from the user in the backend which should not be edited or deleted at any time. For example blocks to insert a Skip to Content link or a Hamburger menu icon. 

One other feature I'd like to see is the possibility to lock certain blocks from editing. Right now it would break the theme layout if users add more columns to the header. I would like to have some more control here, so users can edit the site title and navigation block and add more blocks within the columns, but not break the responsive layout by adding or deleting columns.

Beside that, I'm just waiting until more blocks like the Query or Comments block are available.

---
I have planned to create issues for these points in the Guteberg repo the next days if I don't find preexisting ones. I haven't followed the development closely, so some things mentioned might be already covered :)

ambitious/block-template-parts/front-page/front-page-intro.html
kjellr: > I saw the about.html and contact.html files in the Parisienne theme and thought theme authors can now directly provide block templates for starter content.

Yeah, I think this would be a great improvement, either through "Starter Page Templates" as described above, or through something like what was explored in https://github.com/WordPress/theme-experiments/pull/16.

jffng:
Great feedback, thank you for taking the time to share it!

> It would be nice if theme authors had some Container Block which allows to add semantic markup like `<header>` and `<footer>`. 

Definitely, this is starting to be explored here: https://github.com/WordPress/gutenberg/pull/20218

And regarding the query block: https://github.com/WordPress/gutenberg/pull/20106 

> I have planned to create issues for these points in the Guteberg repo the next days if I don't find preexisting ones. I haven't followed the development closely, so some things mentioned might be already covered :)

💪 This might be a useful project board to find / follow issues related to some of your concerns: https://github.com/WordPress/gutenberg/projects/35

Netzberufler:
Thank you for the links, very helpful :)

jffng:
@Netzberufler I'm going to merge this as I think it's a robust example for demonstrating how the existing theme template hierarchy can be ported to block templates. 

I do think we should think more about how to address what's effectively starter content (https://github.com/WordPress/theme-experiments/pull/21#discussion_r380780371).

dashkevych:
> I also felt that the Group block generates a lot of HTML markup, especially if you just want to wrap smaller elements like the post date.
I was not aware of this. I've checked your demo and I agree with you that it generates a lot of HTML. Is there any plans in a core to address this problem?

Theme HTML markup

Netzberufler:
Thanks so much for merging the theme.

I had some time today to create these two tickets:
https://github.com/WordPress/gutenberg/issues/20375
https://github.com/WordPress/gutenberg/issues/20376

PR #22, opened 2020-02-17
Try/global styles theme

nosolosw:
Creates a barebones theme to highlight how Global Styles would work with themes and provide a testing environment for `edit-site`.


nosolosw:
This is a work in progress (could someone add me as _asignee_ and the _in progress_ label?).

nosolosw:
I've realized that this kind of testing playground/short-lived theme for demo purposes is probably not a good idea to land as part of theme-experiments. I'm closing this in favor of https://github.com/nosolosw/global-styles-theme/pull/5

jffng:
That makes sense, but I hope that a theme taking advantage of the latest global styles features can land here soon.

PR #25, opened 2020-02-24
[Parisienne] Pull in block starter content from dedicated directory

jffng:
Following #16, this PR separates starter content in the Parisienne theme from block template parts — which should conceptually be reserved for aspects of the site that are used across pages, not demo content or page templates.

To test:

2. Check out this branch and clone it into your site's themes directory
3. Install the Gutenberg plugin
4. Enable the "Full Site Editing" experiment in the Gutenberg > Experiments panel.
5. Activate `Parisienne` from within Appearance > Themes.
6. Go to Customizer and hit Publish
7. Visit the site and ensure that the `Home`, `About`, and `Contact` pages have loaded with content in each of them.


parisienne/block-template-parts/header.html
```diff
@@ -1,7 +1,7 @@
 <!-- wp:group {"align":"full"} -->
 <div class="wp-block-group alignfull"><div class="wp-block-group__inner-container"><!-- wp:columns {"align":"full"} -->
-<div class="wp-block-columns alignfull"><!-- wp:column {"width":20} -->
-<div class="wp-block-column" style="flex-basis:20%"><!-- wp:navigation {"className":"is-style-dark"} -->
+<div class="wp-block-columns alignfull"><!-- wp:column {"width":25} -->
```
kjellr: These changes seem unrelated, right? Just want to make sure they're intentional.

jffng: Unrelated! Going to pull them out of this PR.

jffng:
This was a nice thought, but content like this should really exist in a pattern if at all in the theme.

PR #26, opened 2020-02-27
[Twenty Twenty Blocks] Try adding early Global Styles Support

kjellr:
This PR starts to add global styles support to the Twenty Twenty Blocks theme. This is still super-early, so it's pretty limited in its scope right now. It just does the following:

- Adds an `experimental-theme.json` file, with the [3 native Gutenberg variables](https://github.com/WordPress/gutenberg/blob/6f1e87b1e041921dbc9e5a0582dd52920db71fbd/experimental-default-global-styles.json) so far. 
- Swaps out those color values for CSS variables throughout the theme. 

The theme shouldn't look any different on the front end (unless you're using IE at least), but you'll notice the CSS variables when you check the inspector: 

Screen Shot 2020-02-27 at 2 47 06 PM

Screen Shot 2020-02-27 at 2 46 55 PM

Eventually, these colors should show up as editable in [the full-site editing panel](https://github.com/WordPress/gutenberg/pull/20061).

### DIFF
--- a/twentytwenty-blocks/experimental-theme.json
+++ b/twentytwenty-blocks/experimental-theme.json
@@ -1,0 +1,7 @@
+{
+	"color": {
+		"primary": "#cd2653",
+		"background": "#f5efe0",
+		"text": "#000"
+	}
+}
--- a/twentytwenty-blocks/experimental-theme.json
+++ b/twentytwenty-blocks/experimental-theme.json
@@ -7,15 +7,7 @@
 	"typography": {
  		"font-scale": 1.3,
  		"font-size": "18px",
- 		"font-size-heading-1": "64px",
- 		"font-size-heading-2": "48px",
- 		"font-size-heading-3": "40px",
- 		"font-size-heading-4": "32px",
- 		"font-size-heading-5": "24px",
- 		"font-size-heading-6": "18px",
  		"font-weight": 400,
- 		"font-weight-heading": 700,
- 		"line-height": 1.5,
- 		"line-height-heading": 1.25
+ 		"line-height": 1.5
  	}
 }
--- a/twentytwenty-blocks/experimental-theme.json
+++ b/twentytwenty-blocks/experimental-theme.json
@@ -3,5 +3,19 @@
 		"primary": "#cd2653",
 		"background": "#f5efe0",
 		"text": "#000"
-	}
+	},
+	"typography": {
+ 		"font-scale": 1.3,
+ 		"font-size": "18px",
+ 		"font-size-heading-1": "64px",
+ 		"font-size-heading-2": "48px",
+ 		"font-size-heading-3": "40px",
+ 		"font-size-heading-4": "32px",
+ 		"font-size-heading-5": "24px",
+ 		"font-size-heading-6": "18px",
+ 		"font-weight": 400,
+ 		"font-weight-heading": 700,
+ 		"line-height": 1.5,
+ 		"line-height-heading": 1.25
+ 	}
 }
--- a/twentytwenty-blocks/block-styles.css
+++ b/twentytwenty-blocks/block-styles.css
@@ -29,8 +29,8 @@
 }
 
 .is-style-header-nav .wp-block-navigation-link > ul {
-	background: #000;
-	border-color: #000;
+	background: var(--wp--color--text);
+	border-color: var(--wp--color--text);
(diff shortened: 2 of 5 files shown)


jffng:
👏 Which version or branch of GB is this PR developed against?

kjellr:
> Which version or branch of GB is this PR developed against?

Good question! I should've included that above. This should work with Gutenberg 7.6 and/or Gutenberg's `master` branch.

I pushed some additional changes that make this mostly compatible with https://github.com/WordPress/gutenberg/pull/20530 as well though. If you run it using that branch, you'll be able to edit these variables in the beta Site Editor. This is a little crude and experimental, but... this is an experimental repo. 😄 

If you'd like to test that out, here are some rough instructions: 

- Install & activate this branch of the Twenty Twenty Blocks theme. 
- Build & activate https://github.com/WordPress/gutenberg/pull/20530 
- Enable the FSE Experiment under `Gutenberg > Experiments`.
- Go to `Gutenberg > Site Editor (beta)` and play with the Global Styles controls in the sidebar.
- If you'd like, save the changes to `wp_global_styles` via the "Update design" button in the editor.
- Go to the front and see that your changes are applied. It should be _roughly_ correct. 

Here's a quick GIF of that in action: 

global-styles

twentytwenty-blocks/experimental-theme.json
```diff
@@ -0,0 +1,21 @@
+{
+	"color": {
+		"primary": "#cd2653",
+		"background": "#f5efe0",
+		"text": "#000"
+	},
+	"typography": {
```
nosolosw: I've pushed some changes to https://github.com/WordPress/gutenberg/pull/20530 and these are unnecesary now (they're calculated): `font-size-heading-*`, `line-height-heading`, `font-weight-heading`.

kjellr:
Thanks for testing! Let's hold off for now. Once something like WordPress/gutenberg#20530 lands, I think we can merge this in. In the meantime, it makes sense to leave it as a one-off experiment.

twentytwenty-blocks/experimental-theme.json
```diff
@@ -0,0 +1,13 @@
+{
+	"color": {
+		"primary": "#cd2653",
```
nosolosw: As per the current changes at https://github.com/WordPress/gutenberg/pull/20530 primary should be removed.

twentytwenty-blocks/experimental-theme.json
```diff
@@ -0,0 +1,13 @@
...
+		"background": "#f5efe0",
+		"text": "#000"
+	},
+	"typography": {
```
nosolosw: As per feedback, the variable generation mechanism is not going to make it into https://github.com/WordPress/gutenberg/pull/20530 so I have to change the typography variables to something like:

```
	"typography": {
		"font-base": 16,
		"font-scale": 1.2,
		"font-weight-base": 400,
		"font-weight-heading": 400,
		"line-height-base": 1.5,
		"line-height-heading": 1.5
	}
```

nosolosw: Note how we declare base and heading alternatives for weight & line-height.

I reckon this is in flux and it puts so it takes a bit of following up to keep up with the changes. I'm happy to submit these as they happen in the Gutenberg PR if that's helpful.

nosolosw: Note for font-sizes: the current UI proposal by @karmatosed https://github.com/WordPress/gutenberg/issues/21030 only exposes base and scale in the UI.

Hopefully, in a subsequent iteration, we're able to use `calc` or something along those lines in the `theme.json` itself. However, for this PR we're just pushing computation to CSS-land, based on `font-base` and `font-scale` values. See [example](https://github.com/WordPress/gutenberg/pull/20530/files?file-filters%5B%5D=.scss#diff-f2f5efb13e0e16e61312aa9b0c028ae1R17).

nosolosw: I'm pushing at https://github.com/nosolosw/theme-experiments/tree/add/twentytwenty-globalstyles what I have to make this work with the latest version of the Gutenberg PR.

kjellr:
This no longer works, so I'm going to close for now. I'll try to spin up a new PR that follows the [new documentation](https://developer.wordpress.org/block-editor/developers/themes/theme-json/).

PR #32, opened 2020-03-27
add the post comments block to 2019

youknowriad:
This PR adds the recent core/post-comments block to the singular template of 2019-blocks.
The design is not exactly the same as 2019 because it uses the default comments template. 
Potentially, we could add a "comments.php" template to the theme and the block would pick it.


kjellr:
Awesome, looks good. I added a couple small edits: 

- Wrapped the comments in a Group block so they can use the usual Twenty Nineteen comments wrapper classname. 
- Cleaned up some styles to get the width correct. 

Screen Shot 2020-03-27 at 2 36 04 PM

PR #34, opened 2020-04-07
Try adding early Block Patterns support to Twenty Twenty

melchoyce:
This includes one new pattern, based on the Exhibitions grid on https://2020.wordpress.net/. 

See functions.php for changes.


kjellr:
This is very cool! One small suggestion: do you think you could add the block pattern to the existing `twentytwenty-blocks` theme instead? That way the PR will be easier to understand, and we can keep the Twenty Twenty-related improvements consolidated on that theme.

melchoyce:
Sounds good — I'm going to close this PR and open a new one.

PR #35, opened 2020-04-08
[Twenty Twenty Blocks] Try adding early Block Patterns support to the Twenty Twenty Blocks theme

melchoyce:
This includes one new pattern, based on the Exhibitions grid on https://2020.wordpress.net/.

image


twentytwenty-blocks/functions.php
```diff
@@ -307,3 +307,16 @@ function twentytwentyblocks_register_styles() {
 	wp_enqueue_style( 'twentytwentyblocks-style', get_stylesheet_uri(), array(), $theme_version );
 }
 add_action( 'wp_enqueue_scripts', 'twentytwentyblocks_register_styles' );
+
+
+/**
+ * Register Block Patterns.
+ */
+
+register_pattern (
```
jffng, line 316: Let's wrap this function call to be safe / avoid errors if the user is running an older version of Gutenberg.

```php
if ( function_exists( 'register_pattern' ) ) {
  register_pattern (
    ...
  )
}
```

kjellr:
Looks great! Once this incorporates @jffng's suggestion above, I think it's ready to go. 👍

twentytwenty-blocks/functions.php
melchoyce: Pushed an update!

twentytwenty-blocks/functions.php
```diff
@@ -313,10 +313,12 @@ function twentytwentyblocks_register_styles() {
...
-    )
-);
\ No newline at end of file
+if ( function_exists( 'register_pattern' ) ) {
+	register_pattern (
+	    'twentytwenty/exhibitions-pattern',
```
kjellr, line 318:
```suggestion
	    'twentytwenty-blocks/exhibitions-pattern',
```

Super-minor note here, but I think this should match the theme slug.

PR #37, opened 2020-04-10
Add Blockbase theme

kjellr:
This PR adds a new theme called "Blockbase". It's built off the Gutenberg Starter theme, but whereas that theme was built to mimic Gutenberg on the front-end, this theme is built to be a blank slate on both the front and back ends. It features: 

- Basic block-based HTML page templates, with header and footer template parts. 
- The minimum amount of CSS to support the front end, editor, and beta site editor.
- Global Styles support via `theme-experiments.json`. 

## Screenshots: 

Front-end|Full-site Editor
------------|------------
gutenberg test__p=30|wordpress test_wp-admin_admin php_page=gutenberg-edit-site (1)


youknowriad:
I'm not sure if it's the goal of this particular theme but I'd personally want to see a theme without CSS and just relying on Global styles and block styles. Potentially some global reset might be needed for like body margins or things like but it shouldn't exceed a dozen lines of CSS.

kjellr:
> I'm not sure if it's the goal of this particular theme but I'd personally want to see a theme without CSS and just relying on Global styles and block styles. Potentially some global reset might be needed for like body margins or things like but it shouldn't exceed a dozen lines of CSS.

Me too. Eventually, I hope everything in here becomes totally unnecessary. Aside from a general CSS reset, the bulk of the CSS here does two things: 

- It maps the global variables to blocks (This will be handled by Global Styles once it is fully implemented, so I expect this code to disappear soon). 
- It implements default margins, wide/full spacing, and other missing block styles on the front end (This _should_ be provided by Gutenberg eventually, but it isn't today).

youknowriad:
> It maps the global variables to blocks (This will be handled by Global Styles once it is fully implemented, so I expect this code to disappear soon).
> It implements default margins, wide/full spacing, and other missing block styles on the front end (This should be provided by Gutenberg eventually, but it isn't today).

This is great feedback and something we might not be aware of on the Gutenberg repo. I'd appreciate if these things are raised somehow so we can fix them. I guess once we have the initial Global styles implementation we can get into the specifics for each missing customization.

blockbase/css/editor/variables.css
```diff
@@ -0,0 +1,46 @@
...
+	--wp--typography--font-scale: 1.3;
+	--wp--typography--font-family: sans-serif;
+	--wp--typography--font-family-code: monospace;
+	--wp--typography--font-size: "16px";
+	--wp--typography--font-weight: 500;
+	--wp--typography--line-height: 1.7;
+
+	/* Typography Headings */
+	--wp--typography--font-family-heading: sans-serif;
+	--wp--typography--font-size-heading-1: "40px";
```
jffng, line 26: I think `font-size` values need to lose the quotation marks for them to register as valid.

```suggestion
	--wp--typography--font-size-heading-1: 40px;
```

blockbase/style-editor.css
```diff
@@ -0,0 +1,85 @@
...
+.editor-styles-wrapper {
+	background: var(--wp--color--background);
+	color: var(--wp--color--text);
+}
+
+body > *,
+.editor-styles-wrapper > *,
+.editor-styles-wrapper p {
+	font-family: var(--wp--typography--font-family);
+	font-size: var(--wp--typography--font-size);
```
jffng, line 27: Right now this is being overridden in the editor by two higher-precedence style rules injected by Gutenberg: 

Screen Shot 2020-04-10 at 4 26 18 PM

kjellr:
Just a heads up that I tidied this up over the weekend, removing even more unnecessary CSS. Should be ready for another review.

jffng:
Looking good.

I think comments I left in the earlier review are still relevant — mainly the first one, that the font-size variables aren't registering as valid in the editor because of the quotation marks.

jffng:
@kjellr what are your thoughts on this PR? I think it should be brought up-to-date with the latest theme config, or closed.

kjellr:
I think we should close for now. 👍

PR #40, opened 2020-05-30
Add implementation to require Gutenberg in FSE themes.

aristath:
A simple implementation using admin-notices to allow FSE themes to require Gutenberg as a plugin.
The code can be copy-pasted inside a theme directly.

### DIFF
--- a/require-gutenberg/require-gutenberg.php
+++ b/require-gutenberg/require-gutenberg.php
@@ -1,0 +1,329 @@
+<?php
+/**
+ * Plugin Name: Require Gutenberg
+ *
+ * @package WPThemes/Require-Gutenberg-FSE
+ */
+
+if ( ! class_exists( 'WPThemes_Require_Gutenberg' ) ) {
+
+	/**
+	 * Init Gutenberg requirement.
+	 *
+	 * @since 1.0.0
+	 */
+	class WPThemes_Require_Gutenberg {
+
+		/**
+		 * Add hooks for this class.
+		 *
+		 * @access public
+		 *
+		 * @since 1.0.0
+		 *
+		 * @return void
+		 */
+		public function run() {
+
+			// Make sure we have the plugin-checking functions we need.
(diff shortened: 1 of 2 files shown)


require-gutenberg/require-gutenberg.php
```diff
@@ -0,0 +1,329 @@
+<?php
+/**
+ * Plugin Name: Require Gutenberg
```
kjellr, line 3:
```suggestion
 * Require Gutenberg
```

^ Super-minor suggestion, since this the README doesn't suggest including this as a plugin.

aristath: :laughing: you're right, It was just easier to develop as a plugin 'cause I could test with multiple themes that way.

fixed :+1:

aristath:
Hey @jffng!

That's weird... I just pushed a commit on https://github.com/WordPress/theme-experiments/commit/f4c78d73f0b9b4aadabbbce4c5f74fbe9a33b415 and it should help...
No matter what I do I can't replicate this on local or a live server, so hopefully that commit will give us some more info

jffng:
Ah, this is my mistake I think:

```js 
{ 
  install: "plugin", 
  slug: "gutenberg", 
  pluginName: "Gutenberg", 
  errorCode: "folder_exists", 
  errorMessage: "Destination folder already exists." 
}
```

I had previously symlinked my Gutenberg development branch to the plugins directory of the site I was testing on. Removing the symlink solved the issue. 

-----

Where should this live? Does it make sense to include this as an addition to #37?

aristath:
> Where should this live? Does it make sense to include this as an addition to #37?

Ideally it shouldn't be tied to a specific theme, that's why I pushed it on its own folder with a separate readme file.
I'll write a post on `make` and it will link to this repo so that theme-authors can get it and use it for the themes they submit on `.org` (we'll begin accepting FSE themes there soon)

PR #44, opened 2020-06-23
Twenty Twenty Blocks: Try adding new Global Styles Rules

kjellr:
Takes a quick pass at adding Global Styles rules to Twenty Twenty Blocks, as per [the new documentation](https://developer.wordpress.org/block-editor/developers/themes/theme-json/). Essentially an updated version of #26.


twentytwenty-blocks/twentytwenty-styles/editor-style-block.css
```diff
@@ -249,29 +254,30 @@ Inter variable font. Usage:
 .editor-styles-wrapper .wp-block h6 {
 	font-feature-settings: "lnum";
 	font-variant-numeric: lining-nums;
-	font-weight: 700;
+	font-weight: var(--wp--typography--font-weight-heading);
```
jffng, line 257: Have the `--wp--typography--` variables gone away? I don't see them being applied anymore when I activate this theme. Do they need to be converted to presets in the theme config?

twentytwenty-blocks/functions.php
```diff
@@ -52,29 +52,19 @@ function twentytwentyblocks_theme_support()  {
...
 				array(
-					'name'  => __( 'Primary', 'twentytwenty-blocks' ),
-					'slug'  => 'primary',
-					'color' => '#0073AA',
-				),
-				array(
-					'name'  => __( 'Secondary', 'twentytwenty-blocks' ),
-					'slug'  => 'secondary',
-					'color' => '#005177',
+					'name'  => __( 'Background', 'twentytwenty-blocks' ),
```
jffng, line 55: Getting some unexpected behavior when I try to apply custom colors:

Screen Shot 2020-06-25 at 10 23 10 AM

twentytwenty-blocks/twentytwenty-styles/editor-style-block.css
kjellr: Ah, whoops. That's a leftover from the last attempt at this. It needs to be converted to use the new system.

twentytwenty-blocks/functions.php
kjellr: Very weird! 🤔 I'll take another look at that.

twentytwenty-blocks/twentytwenty-styles/editor-style-block.css
kjellr: This should be take care of in fe1e88e.

kjellr: Actually wait — I need to get it fixed in the editor styles too. 👍

twentytwenty-blocks/functions.php
kjellr: Would you mind re-testing this? I can't reproduce anymore.

jffng: Confirmed that this issue is resolved on my end.

twentytwenty-blocks/twentytwenty-styles/editor-style-block.css
kjellr: Ok, should be ok as of a442c52.

kjellr:
@jffng This should be ready for another review. 👍

twentytwenty-blocks/twentytwenty-styles/style.css
```diff
@@ -305,48 +306,47 @@ h6,
...
+	line-height:  var(--wp--preset--line-height--heading);
 	margin: 3.5rem 0 2rem;
 }
 
 h1,
 .heading-size-1 {
-	font-size: 3.6rem;
-	font-weight: 800;
-	line-height: 1.138888889;
+	font-size: var(--wp--preset--font-size--normal-heading-1);
```
jffng, line 317: Where are these `font-size--normal-heading-*` variables coming from?

twentytwenty-blocks/twentytwenty-styles/editor-style-block.css
```diff
@@ -249,37 +254,37 @@ Inter variable font. Usage:
...
+	line-height: var(--wp--preset--line-height--heading);
 	margin: 40px 0 25px;
 }
 
 .editor-post-title__block .editor-post-title__input,
-.editor-styles-wrapper .wp-block h1 {
-	font-size: 36px;
-	font-weight: 800;
+.editor-styles-wrapper h1.wp-block {
+	font-size: 84px;
```
jffng, line 265: Just wondering why we're using px values here when rem values are used in the frontend.

twentytwenty-blocks/twentytwenty-styles/style.css
kjellr: Whoops. I've replaced those in cfeb4dc.

twentytwenty-blocks/twentytwenty-styles/editor-style-block.css
kjellr: In the frontend, Twenty Twenty [modifies the value of `1 rem`](https://github.com/WordPress/theme-experiments/blob/a442c52e37e6003eb66631383813fe2dcef1fffd/twentytwenty-blocks/twentytwenty-styles/style.css#L117) to be `10px` instead of the usual `13px`. Since that doesn't happen in the editor (and can't really, without messing with a lot of other font sizes), we're stuck with pixels for the moment. 

These should really be variables, but I didn't switch to using variables for these sizes because the heading font sizes don't match up with the small, normal, large, + huge sizes. I figured we'd keep things simple for now, and can open up a separate PR improve this.

twentytwenty-blocks/twentytwenty-styles/editor-style-block.css
```diff
@@ -163,21 +168,21 @@ Inter variable font. Usage:
...
 :root .has-primary-color {
-	color: #000;
+	color: var(--wp--preset--color--primary);
 }
 
 :root .has-primary-background-color {
-	background-color: #000;
-	color: #f5efe0;
+	background-color: var(--wp--preset--color--text);
+	color: var(--wp--preset--color--primary);
```
jffng, line 185: There is some strangeness when you apply a background color and text color to buttons with the "Fill" style:

**Editor**
Screen Shot 2020-07-21 at 9 56 20 AM

**Front-end**
Screen Shot 2020-07-21 at 9 56 26 AM

twentytwenty-blocks/twentytwenty-styles/style.css
```diff
@@ -2736,21 +2736,21 @@ h2.entry-title {
 /* CUSTOM COLORS */
 
 :root .has-accent-color {
```
jffng, line 2738: Is the `:root` selector still needed? I noticed that in #45 it wasn't used and still worked.

kjellr: Yeah, these don't get used at all currently — they're overridden by rules assigned to `.editor-styles-wrapper` instead.

twentytwenty-blocks/twentytwenty-styles/editor-style-block.css
kjellr: I had to do a few weird things to get this working: 

1. Rename the "Text" color to "Foreground". Otherwise, the class to indicate that this color was being used was `has-text-color`. This also happens to be the generic classname that's added to blocks when they have _any_ text color assigned. That caused some conflicting styles.
2. I had to rewrite some of the styles, because they seemed to be wrong. 
3. I had to add some extra specificity to get the custom color rules working for buttons. 

It's possible I chopped out a little more code than was necessary here, but it generally seemed to work for me.

twentytwenty-blocks/twentytwenty-styles/editor-style-block.css
```diff
@@ -167,48 +167,34 @@ Inter variable font. Usage:
 
 /* CUSTOM COLORS */
 
-:root .has-accent-color {
+.editor-styles-wrapper .has-primary-color,
```
jffng, line 170: Why do we need to target `.editor-styles-wrapper`? I though that enqueueing via `add_editor_style` was supposed to handle this for us?

kjellr: Ah, that's a good question. The non-blocks version of twenty twenty[ enqueues the editor styles using the old hook](https://core.trac.wordpress.org/browser/trunk/src/wp-content/themes/twentytwenty/functions.php#L391), not via `add_editor_style`. Since we don't do that in the block-based theme here, we should probably strip all these out of the editor stylesheet. They're not necessary anymore. 

I'll leave that for a followup PR though.

PR #45, opened 2020-06-24
Gutenberg Starter Theme Blocks: Try adding new Global Styles Rules

kjellr:
This PR adds basic Global Styles rules to Gutenberg Starter Theme Blocks, as per the [new documentation](https://developer.wordpress.org/block-editor/developers/themes/theme-json/).

### DIFF
--- a/gutenberg-starter-theme-blocks/experimental-theme.json
+++ b/gutenberg-starter-theme-blocks/experimental-theme.json
@@ -86,7 +86,6 @@
 	"core/paragraph": {
 		"styles": {
 			"color": {
-				"background": "var( --wp--preset--color--background )",
 				"link": "var( --wp--preset--color--primary )",
 				"text": "var( --wp--preset--color--foreground )"
 			},
@@ -99,7 +98,6 @@
 	"core/heading/h1": {
 		"styles": {
 			"color": {
-				"background": "var( --wp--preset--color--background )",
 				"link": "var( --wp--preset--color--primary )",
 				"text": "var( --wp--preset--color--foreground )"
 			},
@@ -112,7 +110,6 @@
 	"core/heading/h2": {
 		"styles": {
 			"color": {
-				"background": "var( --wp--preset--color--background )",
 				"link": "var( --wp--preset--color--primary )",
 				"text": "var( --wp--preset--color--foreground )"
 			},
@@ -125,7 +122,6 @@
 	"core/heading/h3": {
 		"styles": {
 			"color": {
-				"background": "var( --wp--preset--color--background )",
 				"link": "var( --wp--preset--color--primary )",
 				"text": "var( --wp--preset--color--foreground )"
 			},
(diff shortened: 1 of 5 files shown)


kjellr:
This should be ready for review. The `experimental-theme.json` file is more or less up to date. I stripped out a lot of the CSS from `style.css` that would be handled by global styles in this iteration (though I'm sure more can be removed too). 

In general, it shouldn't look much different with this patch applied.

gutenberg-starter-theme-blocks/functions.php
```diff
@@ -22,24 +22,24 @@ function gutenberg_starter_theme_blocks_support()  {
 		add_theme_support( 'editor-color-palette', array(
```
jffng, line 22: When I customize colors (e.g. make the paragraph text color Strong Blue), they aren't being applied in the editor right. They do show up in the front-end though.

gutenberg-starter-theme-blocks/experimental-theme.json
```diff
@@ -0,0 +1,204 @@
...
+				"link": "var( --wp--preset--color--primary )",
+				"text": "var( --wp--preset--color--foreground )"
+			},
+			"typography": {
+				"fontSize": "var( --wp--preset--font-size--small )",
+				"lineHeight": "var( --wp--preset--line-height--large )"
+			}
+		}
+	},
+	"core/heading/h6": {
```
jffng, line 164: Just noting the redundancy of some of these styles. It becomes really verbose, and I don't think we want to end up with a json config that's just as long as a theme stylesheet, without a way to compile / generate it.

It would be nice to have a way target multiple blocks with the same styles, like `"core/heading/h*": {`

gutenberg-starter-theme-blocks/functions.php
kjellr: That's odd — I wonder if there's some sort of conflict between the `add_theme_support` colors and the ones in `experimental-theme.json`.

kjellr: Hey would you mind re-testing this? I'm not able to reproduce.

gutenberg-starter-theme-blocks/experimental-theme.json
```diff
@@ -0,0 +1,204 @@
...
+			"color": {
+				"background": "var( --wp--preset--color--background )",
+				"text": "var( --wp--preset--color--foreground )"
+			}
+		}
+	},
+	"core/paragraph": {
+		"styles": {
+			"color": {
+				"background": "var( --wp--preset--color--background )",
```
jffng, line 89: The result of setting this right now is that any paragraph will have a background color applied, e.g. inside the cover block:

Screen Shot 2020-07-21 at 10 15 09 AM

kjellr: Hmm interesting. That's not great. Should we not specify a value by default then?

jffng: Could we set it to `inherit`?

kjellr: I think not specifying a color is what we want here — by default we want all those blocks to just be transparent. I've pushed an update to do that.

PR #49, opened 2020-07-21
Twenty Twenty Blocks: Remove entry-styles-wrapper classes from editor-style-block.css

kjellr:
Followup to #44. @jffng pointed out that `editor-style-block.css` uses the `.editor-styles-wrapper` class throughout it. This is a remnant from the original Twenty Twenty Theme, where the stylesheet was [pulled in](https://core.trac.wordpress.org/browser/trunk/src/wp-content/themes/twentytwenty/functions.php#L391) via the `enqueue_block_editor_assets` hook instead of the usual `add_editor_style`. Since we use `add_editor_style` in the blocks-based version, those `editor-styles-wrapper` classes should be prepended automatically, and are not necessary to use in the stylesheet anymore. 

To test: Compare editor styles on `master` to this branch, and make sure they look and work the same.

### DIFF
--- a/twentytwenty-blocks/functions.php
+++ b/twentytwenty-blocks/functions.php
@@ -299,17 +299,6 @@ function twentytwentyblocks_register_styles() {
 add_action( 'wp_enqueue_scripts', 'twentytwentyblocks_register_styles' );
 
 /**
- * Load styles into Edit Site
- * (This shouldn't be necessary but it seems to be for now.)
- */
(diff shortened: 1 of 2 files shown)


kjellr:
Oh right! I'm not sure we need to enqueue those there anymore. We can either remove that, or we can specify that we only load this in the site editor screen.

kjellr:
@jffng I removed that enqueue entirely, and the editor styles are still working for me. Would you mind giving this another look?

PR #52, opened 2020-09-01
Ambitious: Added Query Loop block and updated styling of post meta and navigation blocks

Netzberufler:
- Made some styling fixes for the latest Gutenberg version
- Made use of the new Query Loop block
- Changed to a flat file structure in block-template-parts folder

Just a minor update to fix things, before I try out global styles :)


kjellr:
👋 Hi @Netzberufler! I gave this a test, and it's mostly working well. I think there's an issue with one of the template parts though: I'm getting this error on the front end for each single page: 

Screen Shot 2020-09-03 at 10 44 30 AM

From a quick look at the code, I think it's looking for an `entry-title` template, but there isn't one with that name:

https://github.com/WordPress/theme-experiments/blob/9eb14305fe9de0038d164b95e02396931c1a7d8b/ambitious/block-template-parts/content-page.html#L5

ambitious/block-template-parts/content-front-page.html
```diff
@@ -31,3 +31,13 @@ <h1 class="has-white-color has-text-color has-text-align-center intro-heading ha
 	</div>
 </div>
 <!-- /wp:group -->
```
jffng, line 33: I think the content above would be great as a block pattern included in the theme, rather than a template part.

ambitious/block-template-parts/content-page.html
```diff
@@ -2,6 +2,7 @@
 <div class="wp-block-group type-page">
 	<div class="wp-block-group__inner-container">
 
+		<!-- wp:template-part {"slug":"entry-title","theme":"ambitious"} /-->
```
jffng, line 5:
```suggestion
		<!-- wp:template-part {"slug":"entry-header","theme":"ambitious"} /-->
```

Netzberufler: Thank you 👍

ambitious/block-template-parts/content-front-page.html
Netzberufler: I agree, I will add that soon. I have already included the pattern for [GT Ambition](https://wordpress.org/themes/gt-ambition/), the existing PHP version of this theme. Works great.

I used the front-page template to see if it could work as some sort of starter content. I'd really like to have some feature to give users a complete page layout rather than a blank page from the start. Patterns are a huge improvement, but still require users to manually discover and insert them. And they work better for smaller sections, not complete pages. WordPress.com has solved this really well with presenting all sorts of block layouts when a new static page is created.

Netzberufler:
> 
> 
> 👋 Hi @Netzberufler! I gave this a test, and it's mostly working well. I think there's an issue with one of the template parts though: I'm getting this error on the front end for each single page:
> 
> Screen Shot 2020-09-03 at 10 44 30 AM
> 
> From a quick look at the code, I think it's looking for an `entry-title` template, but there isn't one with that name:
> 
> https://github.com/WordPress/theme-experiments/blob/9eb14305fe9de0038d164b95e02396931c1a7d8b/ambitious/block-template-parts/content-page.html#L5

Thank you, I missed that. I changed the file structure to make it work in the backend, but it seems subdirectories will be available soon after all :) The ticket was just fixed: https://github.com/WordPress/gutenberg/issues/20375

jffng:
I just tested this as of 9.2.1 and it's looking good. Going to merge and follow up with an update to the readme version, thanks @Netzberufler.

Issue #53, opened 2020-09-16
Add a theme with a magazine-style layout


jffng:
It would be useful to see a block theme that implements a magazine style layout.

Some examples:

- https://bbc.com
- https://verywellhealth.com
- http://www.librarystack.org/

PR #54, opened 2020-09-17
Try a block-based magazine-style theme

kjellr:
Addresses #53. 

Here's a first pass at an example theme using a magazine-style layout. A few caveats: 

- The homepage is basically just a bunch of Latest Post blocks. These should be replaced with Query blocks when they're ready, and would ideally point to different categories/tags. It looked like I couldn't specify categories dynamically in the theme templates for now, so they're all just the most recent posts.
- To make sure there are some posts to view, I bundled some test content, borrowed from the default Gutenberg block patterns. This can be loaded in by installing the theme on a fresh site, visiting the Customizer, and then hitting Publish. 
- I used @aristath's [code snippet here](https://github.com/WordPress/gutenberg/issues/25160#issuecomment-691076081) to implement the alignments and it worked well! (👏 )
- For some reason the editor styles aren't loading? I'll dig back into it later today — I probably made a typo somewhere. 😄 
- To get this out quickly, I just borrowed the overall font + color styles from Twenty Twenty. 
- It still needs a bit of style cleanup. 🙂

The following screenshots are of immediately after a fresh install, with the bundled test content: 

Homepage|Single Post Page
---|---
Screen Shot 2020-09-17 at 11 48 34|Screen Shot 2020-09-17 at 11 49 00


magazine-example/functions.php
```diff
@@ -0,0 +1,133 @@
...
+		add_theme_support( 'custom-line-height' );
+
+		// Add support for experimental link color control.
+		add_theme_support( 'experimental-link-color' );
+
+		// Add support for editor styles.
+		add_theme_support( 'editor-styles' );
+
+		// Enqueue editor styles.
+		add_editor_style( 'editor-editor.css' );
```
jffng, line 34:
```suggestion
		add_editor_style( 'style-editor.css' );
```

kjellr: Perfect example of why it's good to have a second set of eyes on your code. 😄 Thank you!

carolinan:
Hi, is still something that you are working on and want to include in the theme experiments repo?

kjellr:
I'm not sure. 😄 

It's definitely outdated now, and I don't have plans to update it in the short term. I would suggest we close it, but I do think it's an interesting snapshot of an idea for how to execute a Magazine-style site with FSE. We don't really have another example of that in the repo at this point. 

Would anyone else be interested in picking this up (or trying another magazine-style experiment from scratch)?

kjellr:
I'm going to close this one out — it's pretty outdated at this point. 👍

PR #56, opened 2020-10-09
Try a simple two-column landing page theme

kjellr:
@ianstewart pointed me towards this [layout on Carrd](https://a646a4b37deb1e8b.demo.carrd.co) recently, and I wondered how simple it would be to make this as a block-based theme: 

Screen Shot 2020-10-09 at 12 01 48 PM

Turns out... it's pretty simple. 😄 I built a similar one in under an hour, using some of the base from #54. 

## Screenshots

**Homepage** 

Desktop|Mobile
---|---
gutenberg test_(1440x900)|mobile

**Post Page**

Desktop|Mobile
---|---
gutenberg test__p=2524(1440x900) (2)|mobile2


ianstewart:
I especially like this demo because, being familiar with the Site Editor experiment, and how difficult it can be to set up home pages for a site like this with a non-block-based theme … well, it can be really hard doing that. Highlights a big future UX win for users.

jffng (commented):
Very cool! I pushed a commit to update to the latest `experimental-theme.json` spec.

Testing on 9.2 and for some reason an alignfull class is added to the list element of the social icon, causing its width to be 100% and breaking them onto their own line on the front-end: 

Screen Shot 2020-10-21 at 2 46 10 PM

jffng:
Oh and the theme screenshot should be updated!

kjellr:
I just updated the font sizes so they work against Gutenberg `master`. 

> Testing on 9.2 and for some reason an alignfull class is added to the list element of the social icon, causing its width to be 100% and breaking them onto their own line on the front-end:

I'm unable to reproduce on my end. 🤔

kjellr:
I'm going to close this one out, as it's a bit outdated.

PR #57, opened 2020-10-21
Try: Adding a barebones block-based version of the Twenty Twenty-One Theme

kjellr:
An attempt to kick off development for the block-based version of the Twenty Twenty-One theme. 

This generally replicates the look & feel of the original theme. 
**It's meant to be a starting point for more development.**

As of today it includes: 

- Base scaffolding styles (CSS only, no Sass)
- `experimental-theme.json`, populated with defaults from Twenty Twenty-One
- Front Page, Index, Page, and Single templates. 
- Header, Footer Block Template parts
- Block Patterns imported from Twenty Twenty-One
- Block Styles imported from Twenty Twenty-One

Known bugs: 

- It shows a "Template Part Not Found" message when it's loaded from a subdirectory.
- Separator Blocks are doubled up in the editor. 
- Post/Page Titles need styles.
- Block Styles still need some style rules from Twenty Twenty-One: 
	- Latest Posts: Dividers
	- Latest Posts: Borders
	- Editor styles for Social Links: Dark gray color

## Screenshots

Front End|Full Site Editor
---|---
gutenberg test_-(1)|gutenberg test_wp-admin_admin php_page=gutenberg-edit-site-(1)


kjellr:
One other caveat for testing: It looks like folks will need to test against Gutenberg's `master` branch for now. 9.2 appears to be doing something weird to the rendering of the templates — repeating classes and causing some weird bugs on the front end.

carolinan:
I really wish there was even a minimum of testing of how a PR affects FSE before its merged in to Gutenberg.

_nags_ experimental does not mean it should be broken every two weeks.

jffng (commented):
> One other caveat for testing: It looks like folks will need to test against Gutenberg's master branch for now. 9.2 appears to be doing something weird to the rendering of the templates — repeating classes and causing some weird bugs on the front end.

Yeah, this is the front-end running Gutenberg on master: 

Screen Shot 2020-10-21 at 4 05 01 PM

And this is the front-end with 9.2 installed:

Screen Shot 2020-10-21 at 4 15 20 PM

Seems like some kind of parsing error going on: 
Screen Shot 2020-10-21 at 4 15 39 PM

@nosolow do you know if this will be resolved in a minor release?

nosolosw:
I hope so! There's another bug (regression for the code block) that people also want to be fixed. Not sure about the timing but should be soon. Some more context: https://wordpress.slack.com/archives/C02QB2JS7/p1603349339442700?thread_ts=1603311591.436000&cid=C02QB2JS7

mtias:
Having a theme target would also help start building E2E tests to prevent front-end regressions. Excited to see this!

twentytwentyone-blocks/block-template-parts/footer.html
```diff
@@ -0,0 +1,19 @@
+<!-- wp:spacer {"height":70} -->
+<div style="height:70px" aria-hidden="true" class="wp-block-spacer"></div>
```
mtias, line 2: Would you say this should be better expressed as top padding on the `footer` template part if we had such a control?

kjellr: That would certainly be one way to do it — it would come in handy for the header too. At the moment, those template parts are truly just empty wrappers with no settings (other than alignment, I guess). But I could see them being a little more full-featured: padding, background color, etc.

carolinan: An additional, traditional .site-footer class name with styles works too ;)

kjellr:
> It shows a "Template Part Not Found" message when it's loaded from a subdirectory.

If anyone can help figure out why this happens, that would be great. All the other themes in this repo work well if you load them from within `wp-content/themes/theme-experiments/themename`, but this one throws that error. It works fine if it's loaded from `wp-content/themes/themename`. 🤔

twentytwentyone-blocks/block-template-parts/footer.html
mtias: Template parts (at least these two) should be more "robust" — output a semantic tag, have style attributes, and yes, a class :)

It might be nice to conceptualize template parts as `site-` fragments. So by default template parts would output a class of `site-{template-part-name}`.

scruffian:
> If anyone can help figure out why this happens, that would be great. All the other themes in this repo work well if you load them from within wp-content/themes/theme-experiments/themename, but this one throws that error. It works fine if it's loaded from wp-content/themes/themename. 🤔

Fixed here: https://github.com/WordPress/gutenberg/pull/26391

twentytwentyone-blocks/assets/css/blocks.css
```diff
@@ -0,0 +1,171 @@
...
+	.is-style-twentytwentyone-columns-overlap .wp-block-column:nth-child(2n) > p:not(.has-background),
+	.is-style-twentytwentyone-columns-overlap .wp-block-column:nth-child(2n) > h1:not(.has-background),
+	.is-style-twentytwentyone-columns-overlap .wp-block-column:nth-child(2n) > h2:not(.has-background),
+	.is-style-twentytwentyone-columns-overlap .wp-block-column:nth-child(2n) > h3:not(.has-background),
+	.is-style-twentytwentyone-columns-overlap .wp-block-column:nth-child(2n) > h4:not(.has-background),
+	.is-style-twentytwentyone-columns-overlap .wp-block-column:nth-child(2n) > h5:not(.has-background),
+	.is-style-twentytwentyone-columns-overlap .wp-block-column:nth-child(2n) > h6:not(.has-background),
+	.is-style-twentytwentyone-columns-overlap .wp-block-column:nth-child(2n) > ul:not(.has-background),
+	.is-style-twentytwentyone-columns-overlap .wp-block-column:nth-child(2n) > ol:not(.has-background),
+	.is-style-twentytwentyone-columns-overlap .wp-block-column:nth-child(2n) > pre:not(.has-background) {
```
scruffian, line 33: could this be simplified to .is-style-twentytwentyone-columns-overlap .wp-block-column:nth-child(2n) > *:not(.has-background)

kjellr: I don't think so unfortunately — we don't want it to add additional padding to a Group or Cover block for example. The rules are only targeting text-based blocks.

mtias: This is also a good candidate to figure out how to declare it better in theme.json @jorgefilipecosta @nosolosw

scruffian: It seems like this comes up quite a lot. Could we add a class to all text based blocks?

twentytwentyone-blocks/assets/css/blocks.css
```diff
@@ -0,0 +1,171 @@
...
+	}
+}
+
+/*--------------------------------------------------------------
+# Group
+--------------------------------------------------------------*/
+
+.wp-block-group.is-style-twentytwentyone-border {
+	border: 3px solid var(--wp--custom--color--border);
+	padding: var(--global--spacing-vertical);
```
scruffian, line 55: Should this be:
`padding: var(--global--spacing-vertical) var(--global--spacing-horizontal);`

Also these variables don't seem to be set, so when I add this style to a group it looks like this:

Screenshot 2020-10-22 at 15 15 34

twentytwentyone-blocks/assets/css/blocks.css
```diff
@@ -0,0 +1,171 @@
...
+}
+
+/*--------------------------------------------------------------
+# Image
+--------------------------------------------------------------*/
+
+.wp-block-image.alignfull img,
+.wp-block-image.alignwide img {
+	height: auto;
```
scruffian, line 64: I wonder if this should live in the image block itself

twentytwentyone-blocks/assets/css/blocks.css
kjellr: Yeah, themes frequently want to add padding to only-text-based blocks when they're inside of fullwide columns too. A single class would come in handy for that.

twentytwentyone-blocks/assets/css/blocks.css
kjellr: Yeah, I thought it did, but it doesn't seem to. 😕

scruffian: Image blocks don't seem to stretch to full width:
Screenshot 2020-10-22 at 15 21 02

twentytwentyone-blocks/assets/css/blocks.css
```diff
@@ -0,0 +1,171 @@
...
+.wp-block-image.is-style-twentytwentyone-image-frame img {
+	padding: var(--wp--custom--spacing--unit);
+}
+
+/*--------------------------------------------------------------
+# Media & Text
+--------------------------------------------------------------*/
+
+.wp-block-media-text.is-style-twentytwentyone-border {
+	border: 3px solid var(--wp--custom--color--border);
```
scruffian, line 81: Could we make this border apply to everything with this class?:
```
.is-style-twentytwentyone-border {
	border: 3px solid var(--wp--custom--color--border);
}
```

twentytwentyone-blocks/assets/css/blocks.css
kjellr: Ah, good catch! That should variable global-styles.json: `--wp--custom--spacing--vertical` and `--wp--custom--spacing--horizontal`. Fixed in https://github.com/WordPress/theme-experiments/pull/57/commits/894c3516daa6efa564b82af13ac7d2ab9b95975a.

twentytwentyone-blocks/assets/css/blocks.css
kjellr: It only applies to a few blocks, and a couple of them need to be slightly different. The image block for instance needs to be applied to `.is-style-twentytwentyone-border img` instead of just `.is-style-twentytwentyone-border`. And the group block needs some extra padding, so it will still need a `.wp-block-group.is-style-twentytwentyone-border` rule anyway. 

We could declare these all in one place (or define a variable for `3px solid var(--wp--custom--color--border)` at least), but I don't have a strong preference either way.

twentytwentyone-blocks/assets/css/blocks.css
mtias: We don't know what a text based block is, though

scruffian: We could add that to Gutenberg though..

twentytwentyone-blocks/assets/css/blocks.css
kjellr: That's odd — I haven't been able to reproduce. 🤔 I just modified that code because it needs to apply to all images, not just those that are alignwide/alignfull. Mind giving it another try?

kjellr:
I just pushed a few minor updates:

- Updating the main README to include this theme in the list.
- Adding a theme screenshot. 

From the discussion above, it sounds like this is in a decent spot to begin building from. If we have other issues, we can open and explore those as separate issues. 

Would someone mind giving this one more review before merge?

twentytwentyone-blocks/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+To contribute to TT1 development, please read the [contributor's guide](/CONTRIBUTING.md).
+
+## Installation
+
+1. In your admin panel, Install and Activate [the Gutenberg Plugin](https://wordpress.org/plugins/gutenberg/). 
+2. Visit the Gutenberg -> Experiments settings page and check the "Enable Full Site Editing" box. Save your changes. 
+3. Click the 'Code' button on [the main Theme Experiments Repository GitHub page](https://github.com/wordpress/theme-experiments).
+4. Unzip the resulting download. 
+4. Create a new zip file containing only the "twentytwentyone-blocks" folder.
```
jffng, line 17: Because @scruffian's template resolution fix did not make it in 9.2.1, you'll still get a "Template Part Not Found" error if you try to load the theme exists in a subdirectory.

But I don't think this should be a blocker for merging the PR.

jffng (approved):
I think this is a great start @kjellr and sets us up nicely to iterate on the block-based version of TT1 🚢

twentytwentyone-blocks/README.md
kjellr: Cool, we can revise this README once that's merged in. 👍

kjellr:
Thanks for the reviews, everyone. When further issues come up, we can open them as individual issues in this repository. 🚢

twentytwentyone-blocks/assets/css/blocks.css
scruffian: I suggested it here: https://github.com/WordPress/gutenberg/issues/26407

jorgefilipecosta: > This is also a good candidate to figure out how to declare it better in theme.json @jorgefilipecosta @nosolosw

Theme.json is very far from being able to select a selector like this. In order to do so we would need:
- allow theme.json selectors specific to block styles.
	- If we have a block style selector, should the selector be special and accept styles or should it accept settings too? E.g.: in columns with a specific block style, only certain colors are available in the palette?
- Allow nesting of styles e.g.: target group paragraphs inside a column. We don't have nesting, and it seems a theme needs nesting as we can see it is used multiple times in this theme. Should we add a nesting mechanism? If we add nesting, should we have general nesting, or should we also have the option of direct descendent like CSS does with ">"?
- The equivalent of nth-child(2n). Should we have something like that in theme.json? If yes we have two options:
	- Implement an equivalent of where on the run time, we can verify if a block matches the sector and even allow the application of custom settings.
	- Output directly the nth-child(2n) in CSS, so the styles work well but don't do any runtime verification if the block matches a selector or not. This approach would be generic. We would probably have some system to allow any or most of the selectors CSS allows. It may be problematic on mobile as we would probably not be able to apply the styles correctly there.
- The equivalent of not(.has-background). This is a very custom selector. I guess we would not be able to have a runtime way of detecting it (unless we query the dom). Probably the way to address this would be to have a way to use any CSS selector. As referred before, that way would only support styles and not settings.
- Allow outputting of properties we don't have a UI control for yet. Currently, if we don't recognize a theme.json style property, e.g.: marginTop or zIndex, we discard it. I guess we would make theme.json more useful, at least until we don't have more controls if we simply outputted the styles e.g: `zIndex: 2`-> `z-index: 2`.

twentytwentyone-blocks/assets/css/blocks.css
```diff
@@ -0,0 +1,171 @@
...
+	margin-top: 0;
+	margin-bottom: 0;
+}
+
+/*--------------------------------------------------------------
+# Site Title
+--------------------------------------------------------------*/
+
+h1.wp-block-site-title {
+	font-size: calc(1px * var(--wp--preset--font-size--normal));
```
jorgefilipecosta, line 152: Could we use theme.json for the font-szie rule?

twentytwentyone-blocks/functions.php
```diff
@@ -0,0 +1,268 @@
...
+					'slug'     => 'red-to-purple',
+				),
+			)
+		);
+
+		// Add support for responsive embedded content.
+		add_theme_support( 'responsive-embeds' );
+
+		// Add support for custom line height controls.
+		add_theme_support( 'custom-line-height' );
```
jorgefilipecosta, line 225: I guess this supports could now be active with theme.json? Would you be able to confirm @nosolosw ?

twentytwentyone-blocks/style.css
```diff
@@ -0,0 +1,109 @@
...
+	max-width: 360px;
+}
+
+.wp-site-blocks .alignright {
+	float: right;
+	margin-left: 2em;
+	max-width: 360px;
+}
+
+@media screen and (min-width: 1290px) {
```
jorgefilipecosta, line 97: @mtias, @nosolosw Should the v1 of global styles have an equivalent to media queries where we can have different styles depending on screen resolution, or for v1 it is not something we should include?

twentytwentyone-blocks/experimental-theme.json
```diff
@@ -0,0 +1,203 @@
...
+				"spacing": {
+					"customPadding": true,
+					"units": [ "px", "em", "rem", "vh", "vw" ]
+				}
+			},
+			"custom": {
+				"font-primary": "-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif",
+				"font-weight-light": 300,
+				"color": {
+					"primary": "var(--wp--preset--color--dark-gray)",
```
jorgefilipecosta, line 91: The user can change the presets and consequently change the value of the primary is it expected?

twentytwentyone-blocks/experimental-theme.json
```diff
@@ -0,0 +1,203 @@
...
+			"typography": {
+				"fontSize": "var(--wp--preset--font-size--normal)",
+				"lineHeight": "var(--wp--custom--line-height--body)"
+			}
+		}
+	},
+	"core/paragraph": {
+		"styles": {
+			"color": {
+				"text": "var(--wp--custom--color--primary)",
```
jorgefilipecosta, line 128: Would applying "text": "var(--wp--custom--color--primary)", at the global level instead of for each text block also work?

twentytwentyone-blocks/experimental-theme.json
```diff
@@ -0,0 +1,203 @@
...
+			}
+		}
+	},
+	"core/heading/h5": {
+		"styles": {
+			"color": {
+				"text": "var(--wp--custom--color--primary)"
+			},
+			"typography": {
+				"fontSize": "var(--wp--preset--font-size--small)",
```
jorgefilipecosta, line 187: It seems like writing styles in theme.json is much more verbose when compared to CSS, when we need to apply the same style to multiple selectors. Should we have a mechanism in theme.json to target multiple selectors? cc: @mtias, @nosolosw

twentytwentyone-blocks/experimental-theme.json
carolinan: I believe the intention is to prepare the different colors this way in order to help developers or advanced users to change the values?
Similar to how the CSS variables can be used in the PHP version of the theme.

twentytwentyone-blocks/functions.php
```diff
@@ -0,0 +1,268 @@
...
+
+		// Add default posts and comments RSS feed links to head.
+		add_theme_support( 'automatic-feed-links' );
+
+		/*
+		 * Let WordPress manage the document title.
+		 * This theme does not use a hard-coded <title> tag in the document head,
+		 * WordPress will provide it for us.
+		 */
+		add_theme_support( 'title-tag' );
```
carolinan, line 32: This is not needed for FSE themes, the title is already added.

twentytwentyone-blocks/experimental-theme.json
kjellr: > Would applying "text": "var(--wp--custom--color--primary)", at the global level instead of for each text block also work?

It should — I believe when I tried it, it didn't work though so I did this instead. I'd have to dig back into it and double check.

twentytwentyone-blocks/assets/css/blocks.css
mtias: We should not be attempting to replicate that exact CSS code but understand what is it trying to accomplish. Why is the margin necessary? Why the z-index? Can it be absorbed in the way columns lay out content?

twentytwentyone-blocks/style.css
mtias: Not included. Once Gutenberg supports custom values for attributes in responsive breakpoints we can look into it.

twentytwentyone-blocks/experimental-theme.json
```diff
@@ -0,0 +1,203 @@
...
+						"color": "#EEEADD"
+					},
+					{
+						"slug": "white",
+						"color": "#FFFFFF"
+					}
+				]
+			},
+			"typography": {
+				"customFontSize": true,
```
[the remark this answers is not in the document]
kjellr: Yeah, you're probably right. I'll remove.

twentytwentyone-blocks/experimental-theme.json
kjellr: For now yes — that's fine. I initially created these custom variables just to make it easier to migrate the existing color rules over from the original version of Twenty Twenty-One. We should be able to remove these and map to the exact preset palette colors instead in a followup PR.

PR #65, opened 2020-10-29
TT1: Require gutenberg plugin

aristath:
Makes onboarding easier.
Instead of manually requiring the user to install gutenberg, activate, then go to the experiments and activate FSE, this adds an admin notice that does everything automatically.
Should make installing and testing a lot easier and will avoid a lot of questions "why is my site not working" 😉

### DIFF
--- a/twentytwentyone-blocks/functions.php
+++ b/twentytwentyone-blocks/functions.php
@@ -266,3 +266,6 @@ require get_template_directory() . '/inc/block-patterns.php';
 
 // Block Styles.
 require get_template_directory() . '/inc/block-styles.php';
+
+// Require the Gutenberg plugin & experiment.
+require_once get_template_directory() . '/inc/require-gutenberg.php';
--- a/twentytwentyone-blocks/inc/require-gutenberg.php
+++ b/twentytwentyone-blocks/inc/require-gutenberg.php
@@ -68,7 +68,7 @@ if ( ! class_exists( 'WPThemes_Require_Gutenberg' ) ) {
 				return;
 			}
 			?>
-			<div class="notice notice-warning require-gutenberg-notice-wrapper notice-alt<?php echo ( $active_step ) ? ' active-step-' . esc_attr( $active_step ) : ''; ?>">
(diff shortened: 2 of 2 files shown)


kjellr:
Thanks @aristath. This is interesting — on one hand yes it does make things easier, but on the other, it does add a lot of code that isn't technically required in a block-based theme. If folks are looking at this theme as an example of what a block-based theme needs, this may be overkill. 

I'm interested to see what @mtias thinks of this.

carolinan:
I have used it for some time and it does make testing faster. I don't have to prepare anything, I just switch the theme.

aristath:
> on one hand yes it does make things easier, but on the other, it does add a lot of code that isn't technically required in a block-based theme.

I agree, it's a lot of "useless" code. However, until FSE is ready and in core, this does make the process a lot more user-friendly... If FSE was ready in a couple of months I'd say this is not necessary. But considering it might take a while before we get there, I'm hoping user-friendliness will increase ingagement. We can always remove it once FSE is merged in core, this is just a temporary measure 👍

youknowriad:
btw, in https://github.com/WordPress/gutenberg/pull/26500 we're looking at auto-enabling FSE for FSE themes.

aristath:
> btw, in WordPress/gutenberg#26500 we're looking at auto-enabling FSE for FSE themes.

Yep, and that's an awesome improvement. Gutenberg however would still be required to be installed as a plugin and activated... In https://github.com/aristath/q/commit/c382ae5e07ffb21df9c3304058a6093fda2788d0 I added tweaks to account for that auto-activation of FSE, so if https://github.com/WordPress/gutenberg/pull/26500 gets merged we can push the same changes here 👍

aristath:
The code was updated in https://github.com/WordPress/theme-experiments/pull/65/commits/599264af39a309bf5e407be53611ee3b4d3b840f due to https://github.com/WordPress/gutenberg/pull/26500 getting merged.
What it now does:
* If Gutenberg is not installed as a plugin or activated, show a notice requiring its installation/activation
* If the installed Gutenberg plugin does not have a `gutenberg_is_fse_theme` function, then it's a previous version (i.e. not the current trunk) and therefore we need to enable the experiment.

PR #73, opened 2020-11-04
Add: Font size names to 2021 blocks

jorgefilipecosta:
This PR adds font size names to 2021 blocks theme. This change makes the font size picker render the font size names currently everything appears as white space.


kjellr (approved):
This works well, thank you!

PR #74, opened 2020-11-04
Add: Font family presets to 2021 blocks.

jorgefilipecosta:
This PR adds a set of font-family presets so we can take advantage and test the global styles font family picking mechanism.

For now, I'm adding just websafe fonts. I think it would be better If we add fonts that "look good" on this theme. cc: @kjellr, @jasmussen, @carolinan, @aristath  in case you have any font suggestions.

### DIFF
--- a/twentytwentyone-blocks/experimental-theme.json
+++ b/twentytwentyone-blocks/experimental-theme.json
@@ -85,6 +85,29 @@
 						"name": "Gigantic"
 					}
 				],
+				"fontFamilies": [
+					{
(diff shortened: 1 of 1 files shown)


kjellr (approved):
Thanks @jorgefilipecosta! If anything, I'd recommend we leave the monospace stack out for now — the monospace fonts just weren't designed to be used at larger font sizes like what we have in this theme. 

But since none of these are the defaults, I don't see much harm in leaving it in for now.

jorgefilipecosta:
I removed the monospace stack 👍 Feel free to iterate on what font families we provide in the future.
I guess we could now merge this PR (I don't have merge permission).

kjellr:
Excellent thank you!

PR #75, opened 2020-11-04
Twenty Twenty-One Blocks: Add post meta. 

kjellr:
This PR adds post meta to the footer of posts. For reference, here are the original theme styles for these:

Screen Shot 2020-11-04 at 2 12 54 PM

This PR gets somewhat close: 

Screen Shot 2020-11-04 at 2 13 50 PM

- It doesn't include any of the text before these, since the blocks don't support that yet. 
- It keeps the post author text in bold, since this is not yet changeable via Gutenberg or global styles. 

---

The PR adds this to the bottom of all single post pages, and also to each post in the main query loop. While I was in there, I made the post titles clickable, which makes this all work way better in general. 😄 

Example of a post in the main index feed: 

Screen Shot 2020-11-04 at 2 40 43 PM

### DIFF
--- a/twentytwentyone-blocks/assets/css/blocks.css
+++ b/twentytwentyone-blocks/assets/css/blocks.css
@@ -186,7 +186,7 @@ hr[style*="text-align: right"],
 
 hr:not(.is-style-wide):not(.is-style-dots),
 .wp-block-separator:not(.is-style-wide):not(.is-style-dots) {
-	max-width: var(--responsive--aligndefault-width);
+	max-width: var(--wp--custom--responsive--aligndefault-width);
 }
 
 hr.alignwide:not(.is-style-wide):not(.is-style-dots),
--- a/twentytwentyone-blocks/block-templates/index.html
+++ b/twentytwentyone-blocks/block-templates/index.html
@@ -1,6 +1,6 @@
 <!-- wp:template-part {"slug":"header","theme":"twentytwentyone-blocks","align":"full", "tagName":"header","className":"site-header"} /-->
 
-<!-- wp:query {"queryId":3,"query":{"perPage":10,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":""}} -->
+<!-- wp:query {"queryId":1,"query":{"perPage":10,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":""}} -->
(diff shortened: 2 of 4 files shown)


twentytwentyone-blocks/assets/css/blocks.css
```diff
@@ -186,7 +186,7 @@ hr[style*="text-align: right"],
 
 hr:not(.is-style-wide):not(.is-style-dots),
 .wp-block-separator:not(.is-style-wide):not(.is-style-dots) {
-	max-width: var(--responsive--aligndefault-width);
+	max-width: var(--wp--custom--responsive--aligndefault-width);
```
kjellr, line 189: I also fixed this variable name — it was incorrect, and was making the separator above the meta display incorrectly.

carolinan:
The author block does support the text before (byline), was this left blank intentionally?

kjellr:
> The author block does support the text before (byline), was this left blank intentionally?

It was — the byline is only set to show up above the author name, not next to it. I'm going to open a GitHub issue for that (as well as for the ability to un-bold the author name) in the Gutenberg repo today.

twentytwentyone-blocks/block-templates/index.html
```diff
@@ -1,8 +1,38 @@
 <!-- wp:template-part {"slug":"header","theme":"twentytwentyone-blocks","align":"full", "tagName":"header","className":"site-header"} /-->
 
+<!-- wp:query {"queryId":3,"query":{"perPage":10,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":""}} -->
```
ntsekouras, line 3: You can set `1` as this is the first Query instance. 

```suggestion
<!-- wp:query {"queryId":1,"query":{"perPage":10,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":""}} -->
```

ntsekouras: Additionally if you skip the `perPage` property it will search for the reading options from site (Blog pages show at most).

ntsekouras (approved):
Nice work! Now the Query block will work as expected. I've left a couple of minor comments there.

We the template testing I found a bug in PostAuthor block which I'll fix in GB now.

twentytwentyone-blocks/block-templates/index.html
kjellr: Great, I'll remove the perPage property then. 👍 Thanks!

PR #80, opened 2020-11-09
Twenty Twenty-One Blocks: Add max-widths for alignments in the Site Editor

kjellr:
This solves (for now) this issue that I filed over here in the Gutenberg repo: https://github.com/WordPress/gutenberg/issues/26630

Basically, the Site Editor does not provide any max-widths out of the box for default and wide blocks. This PR adds them. We'll need this even when a solution for the upstream issue is in place, since we want to define our own widths here. 

Before|After
---|---
Screen Shot 2020-11-09 at 4 04 38 PM|Screen Shot 2020-11-09 at 4 03 07 PM


aristath (approved):
LGTM :+1:

PR #81, opened 2020-11-10
Try: Define the minimum boilerplate code for a block-based theme

kjellr:
There's just a little bit of boilerplate necessary to build a block-based theme today. This PR tries to collect that all in one place.

```
assets
    alignments-front.css
block-templates
    index.html
experimental-theme.json
functions.php
index.php
style.css
```

Theoretically you could use this today as a base to build a theme from, but I gathered it mostly to start identifying the remaining pieces that (hopefully) Gutenberg can start handling automatically. I'll leave some code comments below. 

Some of the `add_theme_support` declarations and the (mostly) empty `experimental-theme.json` file could technically be removed, but I kept them in for discussion purposes. 

cc @jorgefilipecosta and @nosolosw for your thoughts.


blockbase/assets/alignments-editor.css
```diff
@@ -0,0 +1,15 @@
+/*
```
kjellr, line 1: This file can be removed after https://github.com/WordPress/gutenberg/pull/26853 is merged.

blockbase/assets/alignments-front.css
```diff
@@ -0,0 +1,79 @@
+/*
```
kjellr, line 1: This file would hopefully be removed once Gutenberg provides front-end alignments. https://github.com/WordPress/gutenberg/issues/26633

blockbase/experimental-theme.json
```diff
@@ -0,0 +1,28 @@
...
+			"typography": {
+				"customFontSize": true,
+				"customLineHeight": true,
+				"fontSizes": [ ],
+				"spacing": {
+					"customPadding": true,
+					"units": [ "px", "em", "rem", "vh", "vw" ]
+				}
+			},
+			"custom": {
```
kjellr, line 17: These rules are only used in the alignments stylesheets above, so if those are removed, these could be removed too.

blockbase/functions.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+if ( ! function_exists( 'blockbase_support' ) ) :
+	function blockbase_support()  {
+
+		// Alignwide and alignfull classes in the block editor
+		add_theme_support( 'align-wide' );
```
kjellr, line 7: Should block-based themes opt-into wide/full alignments by default?

blockbase/functions.php
```diff
@@ -0,0 +1,37 @@
...
+
+		// Adding support for core block visual styles.
+		add_theme_support( 'wp-block-styles' );
```
kjellr, line 10: Should block-based themes opt-into the opinionated styles by default? (If not, this should be available as a theme.json setting)

blockbase/functions.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+if ( ! function_exists( 'blockbase_support' ) ) :
+	function blockbase_support()  {
```
kjellr, line 4: Many of these settings should either be activated by default for block-based themes, or available as `theme.json` options:

```
add_theme_support( 'align-wide' );
add_theme_support( 'wp-block-styles' );
add_theme_support( 'responsive-embeds' );
add_theme_support( 'experimental-link-color' );
```

blockbase/functions.php
```diff
@@ -0,0 +1,37 @@
...
+			'assets/alignments-editor.css'
+		) );
+    }
+    add_action( 'after_setup_theme', 'blockbase_support' );
+endif;
+
+/**
+ * Enqueue scripts and styles.
+ */
+function blockbase_scripts() {
```
kjellr, line 33: Just thinking out loud... if we can eliminate the add_theme_support calls above, this file would only exist to enqueue the stylesheets. Would it be possible to enqueue the stylesheets from within `theme.json` instead?

blockbase/functions.php
scruffian: Is there an issue for this in Gutenberg?

kjellr: I just opened https://github.com/WordPress/gutenberg/issues/26901 to discuss.

blockbase/functions.php
kjellr: I've opened an issue here to discuss: https://github.com/WordPress/gutenberg/issues/26902

emptytheme/experimental-theme.json
```diff
@@ -0,0 +1,37 @@
...
+			"color": {
+				"gradients": [ ],
+				"palette": [ ]
+			},
+			"typography": {
+				"customFontSize": true,
+				"customLineHeight": true,
+				"fontFamilies": [ ],
+				"fontSizes": [ ],
+				"fontStyles": [ ],
```
jorgefilipecosta, line 13: I guess in most cases one should not have to define fontStyles, fontWeights, textDecorations, and textTransforms. I guess the future one may define per font family. These API's are very new I would probably not include them yet as they will probably change.

emptytheme/functions.php
```diff
@@ -0,0 +1,38 @@
...
+		add_theme_support( 'align-wide' );
+
+		// Adding support for core block visual styles.
+		add_theme_support( 'wp-block-styles' );
+
+		// Adding support for responsive embedded content.
+		add_theme_support( 'responsive-embeds' );
+
+		// Add support for experimental link color control.
+		add_theme_support( 'experimental-link-color' );
```
jorgefilipecosta, line 16: I guess we should enable link color using the theme.json. global.settings.color.link = true;,

emptytheme/style.css
```diff
@@ -0,0 +1,15 @@
+/*
+Theme Name: Empty Theme
```
jorgefilipecosta, line 2: Should we allow all this header information to be specified in theme.json? @aristath, @mtias, @nosolosw

It seems strange that we force a theme to have a CSS file. I guess a theme with just theme.json should be possible.

scruffian: I was thinking the same. That would be nice.

pattonwebz: The theme directory (and core as well) would need, what I assume, to be a pretty sizable overhaul to do away with the need to have any stylesheet.

emptytheme/experimental-theme.json
kjellr: Yeah, that all makes sense. I've removed for now.

emptytheme/style.css
aristath: Yeah, currently core detects themes from their style.css file. So if we want to allow these to be inside the `theme.json` file, a patch would have to be submitted to core at the same time so that these can be properly used.
I agree though, the stylesheet will in many cases be just a just file used to define the theme.

MaggieCabrera:
We've been testing this PR this morning and found that featured images are not enabled by default 😃 

We could add `add_theme_support( 'post-thumbnails' );` to `functions.php`

The docs say:

> To retain backward compatibility, add_theme_support declarations are retrofit in the proper categories. 
> If a theme uses add_theme_support('disable-custom-colors'), it’ll be the same as set global.settings.color.custom to false. If the experimental-theme.json contains any settings, these will take precedence over the values declared via add_theme_support.

But I tried to set global.settings.post.thumbnails = true on the json and it did nothing. I probably did it wrong.

kjellr:
Thanks! I don't think this option exists yet for `theme.json`, so I've added it to the list in https://github.com/WordPress/gutenberg/issues/26901. 

For now, I've added the traditional `add_theme_support( 'post-thumbnails' );` to `functions.php`.

emptytheme/assets/alignments-front.css
```diff
@@ -0,0 +1,79 @@
...
+	width: var(--wp--custom--width--wide);
+	max-width: 100%;
+	margin-left: auto;
+	margin-right: auto;
+}
+
+.wp-site-blocks .alignfull {
+	transform: translateX(calc(0px - var(--wp--custom--width-horizontal)));
+	width: calc(100% + (2 * var(--wp--custom--width-horizontal)));
+	max-width: calc(100% + (2 * var(--wp--custom--width-horizontal)));
```
MaggieCabrera, line 35: Where is this defined? Should it be `--wp--custom--margin-horizontal`?

Firefox devtools tells me it is not set:

Screenshot 2020-11-18 at 13 27 46

nosolosw: Those are generated from the theme.json declaration (see docs for [free-from custom properties](https://developer.wordpress.org/block-editor/developers/themes/theme-json/#free-form-css-custom-properties)). From what I see in [the file](https://github.com/WordPress/theme-experiments/pull/81/files#diff-fbd7c92cbefc0f9735d61095ae8e3b5a5b42521a1853477566cd54d981d7ba1aR16), the ones that will be generated are: `--wp--custom--width--default`, `--wp--custom--width--wide`, and `--wp--custom--margin--horizontal`.

kjellr: Good catch! That should be `--wp--custom--margin--horizontal`. I've updated that in 5a32027

emptytheme/style.css
jorgefilipecosta: I guess as theme.json involves we should allow just a theme.json theme. But this change would touch many parts of our code and is not something essential. Also as we are I guess most themes will still need at least some lines of CSS code. But once theme.json expands having theme information there seems like a good path.

emptytheme/functions.php
jorgefilipecosta: Hi @kjellr, did you faced any problem enabling link color using theme.json global.settings.color.link = true;?

jorgefilipecosta (commented):
Should we also include a simple homepage template that lists the posts as part of the minimum boilerplate?

kjellr:
> Should we also include a simple homepage template that lists the posts as part of the minimum boilerplate?

Yeah, we could — We'd technically need both an `index.html` and a `singular.html` template to make it all work right now. And since it's more than one template, then a `header.html` template part makes sense, to house the site title. 😄 It starts snowballing from there. 

Without any templates, the theme doesn't technically work by default. So it probably makes sense to add in that baseline at least. I've added them in 2c6d4e1.

emptytheme/functions.php
kjellr: Good call! I just added this in c75159f and it works fine. 👍

primenest (commented):
what would the `emptytheme/assets/block-template-parts` file be used for?

kjellr:
> what would the emptytheme/assets/block-template-parts file be used for?

As of [the last commit](2c6d4e1), that folder contains a reusable [header template part](https://github.com/WordPress/theme-experiments/blob/2c6d4e15f35093f0531259c58471ab06e1d7fe79/emptytheme/block-template-parts/header.html). This is used at the top of the `index.html` and `singular.html` templates.

mintplugins:
@kjellr Do you know if the plans to make the FSE HTML files into PHP files should be reflected here yet?

kjellr:
> Do you know if the plans to make the FSE HTML files into PHP files should be reflected here yet?

Not yet — this PR reflects the current state, not a planned future state.

jffng (commented):
`emptytheme/assets/block-template-parts` is an empty file that I think was accidentally created and checked in?

emptytheme/block-templates/index.html
```diff
@@ -0,0 +1,9 @@
+<!-- wp:template-part {"slug":"header","theme":"emptytheme","tagName":"header"} /-->
```
jorgefilipecosta, line 1: I think we should not force a theme to specify the theme on each place the theme references a template part. The core should automatically assume that the template part being referenced is from that theme.

emptytheme/block-templates/singular.html
```diff
@@ -0,0 +1,5 @@
+<!-- wp:template-part {"slug":"header","theme":"emptytheme","tagName":"header"} /-->
```
jorgefilipecosta, line 1: The template parts are not loading on the edit-site screen but it is probably a bug with FSE.

jorgefilipecosta (approved):
LGTM 👍

kjellr:
> `emptytheme/assets/block-template-parts` is an empty file that I think was accidentally created and checked in?

Good catch! That's removed now.

emptytheme/block-templates/index.html
kjellr: Yeah, definitely. This seems like an unnecessary attribute in this case.

emptytheme/block-templates/singular.html
kjellr: I think that's likely due to [the Gutenberg change noted here](https://github.com/WordPress/gutenberg/issues/27072#issuecomment-733108038). I loaded this up on a fresh site just now and it loads as expected.

kjellr:
We just [discussed this PR in the Block-based Themes chat this week](https://wordpress.slack.com/archives/C02RP4Y3K/p1606922160403400), and decided that it makes sense to merge this PR in. It's been relatively solid the past couple weeks, and I think we hit all of the main discussion points in here. Plus, merging it will make @MaggieCabrera's theme generator easier for folks to use. 🙌

I'll go ahead and do that, and we can continue iterating in additional issues/PRs. Thanks everyone!

PR #86, opened 2020-11-17
TwentyTwentyOne blocks: fix custom spacing support

nosolosw:
Custom spacing was recently made stable, and it lost the `experimental-` prefix.


nosolosw:
Note that updating the theme support name still won't work because there's a pending PR in Gutenberg to fix it https://github.com/WordPress/gutenberg/pull/27045

nosolosw:
cc @jffng @kjellr

jffng (approved):

jffng:
Going to merge this so that the custom spacing support will be restored when the fix is merged to Gutenberg.

PR #90, opened 2020-11-18
TT1 blocks: remove redundant color declarations

jffng:
This PR removes the color settings for heading blocks, because it's already set at the global level. There should be no visual change as a result. 

Is there a good reason to set the heading's text color if it's the same as the global context's text color?


kjellr (approved):
Good call. You're right, there isn't any reason to define it here since the blocks should inherit the global rule anyway. 👍

PR #95, opened 2020-11-25
Empty Theme: Script to generate a personalized starter theme

MaggieCabrera:
Using https://github.com/Automattic/underscores.me/blob/master/plugins/underscoresme-generator/underscoresme-generator.php as a reference I created a script that will generate a zip file with a theme based on `emptytheme` that has the name the user gives to the script.

Run `php new-empty-theme.php` from the root of the repo.

We can expand on this to add further info for style.css such as Author, Description, etc. I only did the name as a starting point.

### DIFF
--- a/new-empty-theme.php
+++ b/new-empty-theme.php
@@ -63,7 +63,7 @@ class Generate_Theme {
 			return $contents;
 		
 		// Special treatment for style.css
-		if ( $filename === 'style.css' ) {
+		if ( 'style.css' === $filename ) {
 			$theme_headers = array(
 				'Theme Name'  => $this->theme['name'],
 				'Text Domain' => $this->theme['slug'],
@@ -74,7 +74,12 @@ class Generate_Theme {
 			}
 
 			$contents = preg_replace( '/\b_s\b/', $this->theme['name'], $contents );
+			return $contents;
+		}
 
+		// Special treatment for functions.php
+		if ( 'functions.php' === $filename ) {
+			$contents = str_replace( 'emptytheme', $this->theme['slug'], $contents );
 			return $contents;
 		}
 	}
--- a/new-empty-theme.php
+++ b/new-empty-theme.php
@@ -7,12 +7,14 @@
 class Generate_Theme {
 
 	protected $theme;
+	protected $old_themename = 'Empty Theme';
+	protected $old_themeslug = 'emptytheme';
 
 	function __construct() {
 
 		$this->theme = array(
-			'name'        => 'Empty Theme',
-			'slug'        => 'emptytheme',
+			'name'        => $this->old_themename,
+			'slug'        => $this->old_themeslug,
 			'uri'         => 'https://github.com/wordpress/theme-experiments/',
 			'author'      => 'Kjell Reigstad',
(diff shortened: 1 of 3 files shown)


new-empty-theme.php
```diff
@@ -0,0 +1,169 @@
...
+
+	function create_zip() {
+
+		if($this->theme['name'] === ''){
+			echo "ABORTING!\n";
+			exit;
+		}
+
+		$zip = new ZipArchive;
+		$zip_filename = sprintf( '%s.zip', md5( print_r( $this->theme['name'], true ) ) );
```
jffng, line 29: This results in a zip with a name as an md5 hash e.g. `098f6bcd4621d373cade4e832627b4f6.zip`. Is this intentional?

jffng (commented):
Looking good and working well for me!

We should add some documentation to the readme about the php requirement and how to use this.

new-empty-theme.php
mendezcode: I think the slug would be better suited for the theme name, e.g. `my-theme.zip` instead of the MD5.

scruffian:
Is it more convenient for this to generate a zip or a directory?

new-empty-theme.php
MaggieCabrera: yeah, it's the approach by the original underscores' script but we can change this definitely

MaggieCabrera:
> Is it more convenient for this to generate a zip or a directory?

I am not 100% sure of how we want to go with this script and that's why I kept the approach of the underscores original code. And also why I didn't add any extra info for the style.css.

I think this is a good starting point and the next steps would be:

- [ ] Changing from generating a zip file to just a folder
- [ ] Ask the user for the author name, theme URL, and description. (Do we have a default in case the user doesn't fill one of these?) 
- [ ] Add documentation for this script in the README

Does that sound good? Anything else we could add?

carolinan:
So happy to see work on this starting 🎉

MaggieCabrera:
This is ready for another review :D

new-empty-theme.php
```diff
@@ -0,0 +1,197 @@
...
+		echo "Author: ";
+		$input = fopen ("php://stdin","r");
+		$this->theme['author'] = trim(fgets($input));
+
+		echo "Theme URI: ";
+		$input = fopen ("php://stdin","r");
+		$this->theme['uri'] = trim(fgets($input));
+	}
+
+	function create_zip() {
```
scruffian, line 51: We should probably change the name of this function!

scruffian (approved):
LGTM

scruffian:
In style.css I still see this:

`Empty Theme is distributed under the terms of the GNU GPL.`

scruffian:
The block template parts also have the theme slug in them.

MaggieCabrera:
> The block template parts also have the theme slug in them.

I think the theme slug should be removed from the template parts like Jorge says here: https://github.com/WordPress/theme-experiments/pull/81#discussion_r531186733

scruffian (approved):
LGTM

kjellr:
🎉 Thanks for this, @MaggieCabrera!

Issue #82, opened 2020-11-12
Twenty Twenty-One Blocks: Migrate over default block styles as theme.json settings.


kjellr:
Many (most) blocks in Twenty Twenty-One have custom styles applied. We should work towards making the block-based version mimic those styles. 

It's ok to migrate over a little CSS for transitions and other extraneous CSS, but basic styles like font specifications should be defined as `theme.json` settings whenever possible. 

This is the full list of blocks that have stylesheets in Twenty Twenty-One. They'll need to be individually audited to see which how much can be migrated over today. Some may already be complete (headings and separator styles for example), so we can check those off after we double check. 

- [ ] [audio](https://github.com/WordPress/twentytwentyone/tree/trunk/assets/sass/05-blocks/audio)
- [ ] [button](https://github.com/WordPress/twentytwentyone/tree/trunk/assets/sass/05-blocks/button)
- [ ] [code](https://github.com/WordPress/twentytwentyone/tree/trunk/assets/sass/05-blocks/code)
- [ ] [columns](https://github.com/WordPress/twentytwentyone/tree/trunk/assets/sass/05-blocks/columns)
- [ ] [cover](https://github.com/WordPress/twentytwentyone/tree/trunk/assets/sass/05-blocks/cover)
- [ ] [file](https://github.com/WordPress/twentytwentyone/tree/trunk/assets/sass/05-blocks/file)
- [ ] [gallery](https://github.com/WordPress/twentytwentyone/tree/trunk/assets/sass/05-blocks/gallery)
- [ ] [group](https://github.com/WordPress/twentytwentyone/tree/trunk/assets/sass/05-blocks/group)
- [ ] [heading](https://github.com/WordPress/twentytwentyone/tree/trunk/assets/sass/05-blocks/heading)
- [ ] [html](https://github.com/WordPress/twentytwentyone/tree/trunk/assets/sass/05-blocks/html)
- [ ] [image](https://github.com/WordPress/twentytwentyone/tree/trunk/assets/sass/05-blocks/image)
- [ ] [latest-comments](https://github.com/WordPress/twentytwentyone/tree/trunk/assets/sass/05-blocks/latest-comments)
- [ ] [latest-posts](https://github.com/WordPress/twentytwentyone/tree/trunk/assets/sass/05-blocks/latest-posts)
- [ ] [list](https://github.com/WordPress/twentytwentyone/tree/trunk/assets/sass/05-blocks/list)
[...]

PR #103, opened 2020-11-30
Twenty Twenty-One Blocks: Migrate Image Block styles

### LINKED ISSUE #92: Twenty Twenty-One Blocks: Migrate Image Block styles
Check that the styles for the image block mirror the ones in TT1. Define as much as possible of the styles within theme.json

Part of #82

MaggieCabrera:
Migrated styles for Image Block.

There are issues that affect this block that would be solved by https://github.com/WordPress/gutenberg/issues/26633 and https://github.com/WordPress/gutenberg/issues/27116

Broken out of #82

Closes #92

### DIFF
--- a/twentytwentyone-blocks/assets/css/blocks.css
+++ b/twentytwentyone-blocks/assets/css/blocks.css
@@ -59,9 +59,21 @@
 # Image
 --------------------------------------------------------------*/
 
-.wp-block-image img,
 .wp-block-image img {
 	height: auto;
+}
+
+.wp-block-image figcaption{
+	text-align: center;
+	color: var(--wp--preset--color--dark-gray);
+	font-size: var(--wp--preset--font-size--extra-small);
+	line-height: var(--wp--custom--line-height--body);
+	margin-top: calc(0.5 * var(--wp--custom--spacing--unit));
+	margin-bottom: var(--wp--custom--spacing--unit);
+}
+
+.wp-block-image a:focus img {
+	outline-offset: 2px;
 }
 
 .wp-block-image.is-style-twentytwentyone-border img,
--- a/twentytwentyone-blocks/style.css
+++ b/twentytwentyone-blocks/style.css
@@ -103,7 +103,7 @@ body {
 
 	.wp-site-blocks .alignfull {
 		transform: translateX(0px);
-		width: 100% + var(--wp--custom--spacing--horizontal));
+		width: 100% + var(--wp--custom--spacing--horizontal);
 		max-width: calc(100% + var(--wp--custom--spacing--horizontal));
 	}
 }


MaggieCabrera:
I have found issues with alignments that affect the image block but require a separate PR altogether.

twentytwentyone-blocks/style.css
```diff
@@ -103,7 +103,7 @@ body {
 
 	.wp-site-blocks .alignfull {
 		transform: translateX(0px);
-		width: 100% + var(--wp--custom--spacing--horizontal));
+		width: 100% + var(--wp--custom--spacing--horizontal);
```
jffng, line 106: What does this rule mean? The width of the block should be 100% of its container plus some horizontal spacing? 

Also I didn't know you could do a math operation outside of calc. The browser does appear to compute it and an alignfull image appears correctly in my testing. But setting this to `100%` appears to have the same effect.

```suggestion
		width: 100%;
```

MaggieCabrera: yeah, I just saw that the extra parenthesis was breaking the alignfull images and that was a quick fix but I'm going to make a separate PR (https://github.com/WordPress/theme-experiments/issues/104) for the alignments since they are not up to date with TT1's. I just left this here as a quick fix in the meantime

twentytwentyone-blocks/assets/css/blocks.css
```diff
@@ -59,11 +59,23 @@
...
+.wp-block-image figcaption{
+	text-align: center;
+	color: var(--wp--preset--color--dark-gray);
+	font-size: var(--wp--preset--font-size--extra-small);
+	line-height: var(--wp--custom--line-height--body);
+	margin-top: calc(0.5 * var(--wp--custom--spacing--unit));
+	margin-bottom: var(--wp--custom--spacing--unit);
+}
+
+.wp-block-image a:focus img {
```
kjellr, line 75: I think we can leave this out until we have a resolution to https://github.com/WordPress/gutenberg/issues/27075 happens. We don't currently have any styles in the heme that would add an `outline` on focus anyway.

twentytwentyone-blocks/assets/css/blocks.css
```diff
@@ -59,11 +59,23 @@
 # Image
 --------------------------------------------------------------*/
 
-.wp-block-image img,
 .wp-block-image img {
 	height: auto;
 }
 
+.wp-block-image figcaption{
```
kjellr, line 66: We _could_ add this here, but I wonder if a better route would be to open a Gutenberg issue to add caption styles to `theme.json`. I can see this being something we'd want to customize in other blocks too.

scruffian: Connected to https://github.com/WordPress/gutenberg/issues/12997

MaggieCabrera: Mmmh, I see. How much should be managed by the json? Looking at these styles I'm inclined to say everything but the margins. But that's maybe because I'm not too clear on how `theme.json` should handle these spacing variables that  we use for our themes.

scruffian: I think the margins could be part of the block settings too, using custom spacing.

kjellr: I think the margins may end up being handled by JSON too eventually.

MaggieCabrera:
I'm closing this PR in favor of https://github.com/WordPress/gutenberg/issues/27480, https://github.com/WordPress/gutenberg/issues/27315 and https://github.com/WordPress/gutenberg/issues/27075

Issue #94, opened 2020-11-25
TT1 Blocks: Migrate CSS for code block


scruffian:
Broken out of https://github.com/WordPress/theme-experiments/issues/82

The code is:

```
.wp-block-code {
	border-color: var(--global--color-border);
	border-radius: 0;
	border-style: solid;
	border-width: 0.1rem;
	padding: var(--global--spacing-unit);

	code {
		font-size: var(--global--font-size-xs);
		white-space: pre;
		overflow-x: auto;
		display: block;
	}
}
```

- The border settings will need to be added to Global Styles.
- IMO the white space and overflow settings can be removed to just use the ones that the block has
- display block should probably be added to the block too
- a padding control should also be added to the block

pbking:
Padding control has been added (see linked Gutenberg PR).
Font Size and default padding have been added to theme.json

BORDER is another matter... there's a [WIP](WordPress/gutenberg#25791) that is adding border RADIUS but in the way that the rest of border will follow; That code needs to land before we can make the CODE's border customizable.
I'm unsure what the timeline for that might be so I've supplied CSS to style the border as was done in TT1.  Not sure if that will stay in.

PR #113, opened 2020-12-08
Migrate CSS for Code block for TT1-block theme

pbking:
Change to support: https://github.com/WordPress/theme-experiments/issues/94

Reference [TwentyTwentyOne's styles](As expressed by TT1's CODE block styles: https://github.com/WordPress/twentytwentyone/blob/trunk/assets/sass/05-blocks/code/_style.scss) for design target.

image

* Supply default padding values for CODE block
* Supply default font size attribute to code block
* Supply default boarder attributes to code block

Note: There were previously dependencies on changes to Gutenberg to allow some of these changes.  The actual changes made were different from the original dependencies but the result was the same.


carolinan:
There is a problem with the text color of the code block, in Twenty Twenty-One, it is being overwritten by the editor styles.
It would be useful to be able to set the text color.
_Probably a lot more useful than being able to set both padding and border radius._

Trac ticket:
https://core.trac.wordpress.org/ticket/51985

scruffian:
As @carolinan suggested we should be able to add support for custom colors to this block like we have for others.

My preference would be to not ship the CSS for borders in the PR but wait for support to be added to Gutenberg.

pbking:
This has been updated to depend completely on Global Styles for border styling.  With this (and its dependent change) the code block for TT1-blocks reaches feature parity with TwentyTwentyOne.

image

carolinan:
I removed the font size because in Twenty Twenty-One the default size of 20px is used.

Issue #131, opened 2020-12-09
TT1 Blocks: Table Block: Migrate CSS to theme.json


scruffian:
The code is [here](https://github.com/WordPress/twentytwentyone/tree/trunk/assets/sass/05-blocks/table).

From #82

jffng:
Opened [this PR](https://github.com/WordPress/gutenberg/pull/27628) to address part the border-collapse discrepancy between the editor and front-end: 

Editor | Front-end
------- | --------
<img width="636" alt="Screen Shot 2020-12-09 at 5 38 02 PM" src="https://user-images.githubusercontent.com/5375500/101697169-50578700-3a45-11eb-8553-2f161fe62328.png"> | <img width="624" alt="Screen Shot 2020-12-09 at 5 37 57 PM" src="https://user-images.githubusercontent.com/5375500/101697175-53527780-3a45-11eb-800e-a88d03c4e80d.png">

PR #133, opened 2020-12-09
TT1 Blocks: migrate table block styles

jffng:
This PR adds the CSS needed to match the table block styles from TT1, addressing part of #131.

It does not address the calendar block which is included in TT1's table block styles but I think should be addressed in another PR. 

I would test the following use cases:
 
- With and without a heading 
- With stripes style variation
- With a background color

Note that this PR requires https://github.com/WordPress/gutenberg/pull/27628 to look fully correct.

### DIFF
--- a/twentytwentyone-blocks/assets/css/blocks.css
+++ b/twentytwentyone-blocks/assets/css/blocks.css
@@ -244,3 +244,15 @@ h1.wp-block-site-title a:not(:hover):not(:focus):not(:active) {
 .wp-block-social-links.is-style-twentytwentyone-social-icons-color .wp-social-link {
 	background: none;
 }
+
+/*--------------------------------------------------------------
+# Table Block
+--------------------------------------------------------------*/
+
+.wp-block-table {
+	min-width: 240px;
+}
+
+.wp-block-table th, .wp-block-table td {
+	border-width: 0;
+}
--- a/twentytwentyone-blocks/assets/css/blocks.css
+++ b/twentytwentyone-blocks/assets/css/blocks.css
@@ -249,10 +249,6 @@ h1.wp-block-site-title a:not(:hover):not(:focus):not(:active) {
 # Table Block
 --------------------------------------------------------------*/
 
-.wp-block-table {
-	min-width: 240px;
-}
-
-.wp-block-table th, .wp-block-table td {
+.wp-block-table.is-style-stripes td {
 	border-width: 0;
 }


twentytwentyone-blocks/assets/css/blocks.css
```diff
@@ -243,4 +243,16 @@ h1.wp-block-site-title a:not(:hover):not(:focus):not(:active) {
...
+/*--------------------------------------------------------------
+# Table Block
+--------------------------------------------------------------*/
+
+.wp-block-table {
+	min-width: 240px;
+}
+
+.wp-block-table th, .wp-block-table td {
+	border-width: 0;
```
jffng, line 257: Now that I see it here, I wonder if we should be able to set the border-width of the table block via theme.json?

scruffian: Yeah I think that would make sense

twentytwentyone-blocks/assets/css/blocks.css
```diff
@@ -243,4 +243,16 @@ h1.wp-block-site-title a:not(:hover):not(:focus):not(:active) {
...
 .wp-block-social-links.is-style-twentytwentyone-social-icons-color .wp-social-link {
 	background: none;
+}
+
+/*--------------------------------------------------------------
+# Table Block
+--------------------------------------------------------------*/
+
+.wp-block-table {
+	min-width: 240px;
```
scruffian, line 253: I wonder if we still need this. @kjellr what do you think?

kjellr: Yeah, probably fine to remove. Gutenberg doesn't set a min-width itself, but I think that's ok. 👍

twentytwentyone-blocks/assets/css/blocks.css
jffng: https://github.com/WordPress/gutenberg/issues/27651

jffng:
With the outstanding Gutenberg issue and PR, this becomes a one line change. I think it should be fine to land or we can put it in blocked until those Gutenberg changes are addressed.

scruffian:
Since we won't need this change once https://github.com/WordPress/gutenberg/issues/27651 lands then I think Blocked makes more sense.

carolinan:
Besides the table caption size, and the text colors breaking in different ways depending on if TT1 blocks or Twenty Twenty-One is active, I do not see what needs to be changed here?
Can this be closed?

left: TT1 Blocks. Right: Twenty Twenty-One
image

carolinan:
Note: The above screenshot is without the PR applied.

scruffian:
Yeah, I think we can close this.

PR #134, opened 2020-12-10
Empty Theme: Fixed alignfull rule causing a horizontal scroll

MaggieCabrera:
Our default alignment rules were causing a horizontal scroll on fully aligned images. This PR changes the width value to fix this.

**Before:**

Screenshot 2020-12-10 at 10 01 43

**After:** 

Screenshot 2020-12-10 at 10 01 52

### DIFF
--- a/emptytheme/assets/alignments-front.css
+++ b/emptytheme/assets/alignments-front.css
@@ -74,6 +74,6 @@ body {
 	.wp-site-blocks .alignfull {
 		transform: translateX(0px);
 		width: 100%;
-		max-width: calc(100% + var(--wp--custom--margin--horizontal));
+		max-width: 100%;
 	}
 }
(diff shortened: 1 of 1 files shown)


emptytheme/assets/alignments-front.css
```diff
@@ -73,7 +73,7 @@ body {
 
 	.wp-site-blocks .alignfull {
 		transform: translateX(0px);
-		width: 100% + var(--wp--custom--margin--horizontal));
+		width: 100%;
 		max-width: calc(100% + var(--wp--custom--margin--horizontal));
```
kjellr, line 77: Thanks! That looks like it was a typo, since it wasn't wrapped in `calc()` anyway. 😄 

We should do the same to the `max-width`, and then this should be good to go. 

```suggestion
		max-width: 100%;
```

kjellr (approved):

PR #136, opened 2020-12-10
Empty Theme: Fix custom spacing entry in theme.json

kjellr:
This was previously declared under "typography", which is incorrect. 😄 

This PR moves it up a level, so you should see it properly registered and reflected in the UI:

Screen Shot 2020-12-10 at 10 00 46 AM

### DIFF
--- a/emptytheme/experimental-theme.json
+++ b/emptytheme/experimental-theme.json
@@ -6,13 +6,13 @@
 				"link": true,
 				"palette": [ ]
 			},
+			"spacing": {
+				"customPadding": true
+			},
 			"typography": {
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
Should this show up on the global styles sidebar too? Or maybe that's not a thing

kjellr:
It'll show up on individual block sidebars now, but no, I don't think this exists in any global styles sidebars yet.

MaggieCabrera (approved):
This looks good to me 🚢

scruffian (approved):
I noticed this too but I forgot to change it here. Thank you for fixing!

PR #137, opened 2020-12-10
Migrate button styles into TT1-blocks

pbking:
#### Adds color & radius configs for button
Adds the expected color and border default values for the theme based
on the 2020 theme and expected changes to Gutenberg.


carolinan:
Is the border **size** here the border radius? I wonder if that could be renamed to be more clear 🤔 

-The buttons border thickness/width is 3px.

PR #139, opened 2020-12-14
TT1 blocks: move spacing section to proper place

nosolosw:
This PR moves the spacing section up one hierarchy level, as it lives alongside the typography and colors sections.

It also removes the `customUnits`. They aren't different from the defaults, hence are unnecessary.


scruffian:
Thanks for fixing this. I wonder if we'll need to add `%` as a custom unit but we can add that back if needed.

PR #143, opened 2020-12-15
Include templates in conversion

pbking:
Previously, themes created using this script would need to have template slug values changed manually for the "stock" content to show up.

This change includes .html files in the change from oldslug > newslug

Tested by creating a new theme using script and loading the theme.  Basic content was showing:

image

### DIFF
--- a/new-empty-theme.php
+++ b/new-empty-theme.php
@@ -86,7 +86,7 @@ class Generate_Theme {
 	function replace_theme_name($contents, $filename) {
 
 		// Replace only text files, skip png's and other stuff.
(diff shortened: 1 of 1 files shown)


MaggieCabrera (approved):
I thought we were going to remove the slug from the templates (https://github.com/WordPress/theme-experiments/pull/81#discussion_r531186733) but it still makes sense to have this in place just in case. Thanks for adding this.

kjellr:
👍 Agreed, makes sense to include this until that change is made. Thanks!

PR #144, opened 2020-12-16
Fix alignment rules

scruffian:
This looks like a typo.

### DIFF
--- a/twentytwentyone-blocks/style.css
+++ b/twentytwentyone-blocks/style.css
@@ -103,7 +103,7 @@ body {
 
 	.wp-site-blocks .alignfull {
 		transform: translateX(0px);
-		width: calc(100% + var(--wp--custom--spacing--horizontal));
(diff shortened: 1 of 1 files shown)


twentytwentyone-blocks/style.css
```diff
@@ -103,7 +103,7 @@ body {
 
 	.wp-site-blocks .alignfull {
 		transform: translateX(0px);
-		width: 100% + var(--wp--custom--spacing--horizontal));
+		width: calc(100% + var(--wp--custom--spacing--horizontal));
```
kjellr, line 106: If I recall from #134, these should just be `100%` right? I don't think the custom spacing was necessary.

scruffian: Yeah that would be better actually, I'll change it...

scruffian: Done

Issue #140, opened 2020-12-15
Rename Twenty Twenty-One Blocks and prepare it for inclusion in the theme directory


carolinan:
It has been brought up that it may be easier for people to test and contribute to full site editing if Twenty Twenty-One blocks 
was available in the WordPress theme directory.

But if we add a second theme called Twenty Twenty-One to the directory, there is a real risk that users
downloads the wrong theme.
To avoid confusion about the differences between Twenty Twenty-One and Twenty Twenty-One Blocks, the full site editing
theme needs to be renamed.

One option would be to use the short version of the name that we as developers have been using: **TT1**, or **TT1 Blocks**.
A screenshot to further differentiate the themes would also be needed.

There are no technical difficulties with adding the theme to the directory, as long as a person with access to the https://wordpress.org/themes/author/wordpressdotorg/ account is available to do the theme uploads.

jffng:
> One option would be to use the short version of the name that we as developers have been using: TT1, or TT1 Blocks.

I like TT1. It's simpler, like the block-based version of the theme.

> There are no technical difficulties with adding the theme to the directory, as long as a person with access to the https://wordpress.org/themes/author/wordpressdotorg/ account is available to do the theme uploads.

We can get started with that process of uploading a blank theme 🚀

kjellr:
Confused people coming across this in search is definitely a concern, and a name change might help with that. 

Let's chat about this in [tomorrow's Block-based Themes meeting](https://make.wordpress.org/themes/2020/12/15/block-based-themes-meeting-agenda-december-16/). 👍

carlomanf:
I'm just surprised no one brought up that there might be confusion between the extremely similar "Twenty Twenty-One" and "Twenty Twenty" but I guess it's too late for that.

carolinan:
> I'm just surprised no one brought up that there might be confusion between the extremely similar "Twenty Twenty-One" and "Twenty Twenty" but I guess it's too late for that.

The default theme names are set, they are all part of a series and they are easy to recognize because of this.

The squad did briefly discuss using a different name also because Twenty Twenty-One _-and upcoming theme names-_ are very long. But it was never brought up with Matt.

carolinan:
- [ ] Update screenshot
- [ ] Update the remaning `@subpackage`
- [ ] Add file header to functions.php

Before release:
Add templates for 

- [ ] search
- [ ]  archive

kjellr:
[...]

PR #145, opened 2020-12-16
Rename theme to TT1 Blocks

jffng:
This PR addresses the name change part of #140.

Based on the discussion in the recent [block-based themes meeting](https://make.wordpress.org/themes/2020/12/16/block-themes-meeting-notes-for-16-december-2020/), folks seemed okay with the name "TT1 Blocks". 

Note that we will need to follow up with a Gutenberg PR to address some tests there: https://github.com/WordPress/gutenberg/search?q=twentytwentyone-blocks

**To test**
- Clone this repository into a fresh site's themes directory
- Install and activate Gutenberg 
- Ensure the site and site editor load as expected

One question — could we update the theme screenshot to somehow visually represent that this is not the same theme as Twenty Twenty-One?


tt1-blocks/inc/block-styles.php
```diff
@@ -68,7 +68,7 @@ function twenty_twenty_one_register_block_styles() {
 			'core/latest-posts',
 			array(
 				'name'  => 'twentytwentyone-latest-posts-dividers',
```
jffng, line 70: I left these as since the pattern is a Twenty Twenty-One pattern...

tt1-blocks/block-templates/front-page.html
```diff
@@ -0,0 +1,5 @@
+<!-- wp:template-part {"slug":"header","theme":"tt1-blocks","align":"full", "tagName":"header","className":"site-header"} /-->
```
scruffian, line 1: Is this a new file?

scruffian: Wait I see its deleted below... weird.

tt1-blocks/inc/block-styles.php
kjellr: I left these originally, and I think it's probably ok to keep as is here too. The benefit is that if someone were to switch from the original Twenty Twenty-One to this one, any blocks using those custom styles would still work, since this theme ships with CSS that targets the same classnames.

scruffian (approved):
LGTM

kjellr:
We'll also want to change the screenshot probably, though we can handle that in a separate PR.

tt1-blocks/block-templates/front-page.html
jffng: That is weird, I wonder why github didn't catch that it was a file rename.

carolinan:
We have confirmation that the theme slug is available.

tt1-blocks/functions.php
```diff
@@ -12,14 +12,14 @@
 	 *
 	 * @return void
 	 */
-	function twenty_twenty_one_blocks_setup() {
+	function tt1_blocks_setup() {
 		/*
 		 * Make theme available for translation.
 		 * Translations can be filed in the /languages/ directory.
 		 * If you're building a theme based on Twenty Twenty-One, use a find and replace
```
carolinan, line 19:
```suggestion
		 * If you're building a theme based on TT1 Blocks, use a find and replace
```

carolinan, line 19:
```suggestion
		 * If you're building a theme based on TT1 Blocks, use a find and replace
```

ntsekouras: If everything else seems fine and since is already approved, @carolinan  do you mind committing your suggestion so as we can merge? GB changes PR is ready as well and will be merged right after this one.

tt1-blocks/inc/block-patterns.php
```diff
@@ -0,0 +1,120 @@
+<?php
+/**
+ * Block Patterns
+ *
+ * @link https://developer.wordpress.org/reference/functions/register_block_pattern/
+ * @link https://developer.wordpress.org/reference/functions/register_block_pattern_category/
+ *
+ * @package WordPress
+ * @subpackage Twenty_Twenty_One
```
carolinan, line 9:
```suggestion
 * @subpackage TT1_Blocks
```

tt1-blocks/functions.php
carolinan: Just need to test run it quickly.

carolinan:
With Gutenberg version 9.5.2, the template parts in the site editor are not loading, I only see the spinner.
Even if I make changes to the template parts and save them, the spinner is showing,.
No visible problems on the front.

I am using Windows 10, Firefox version 84.0, Chrome version 87.0.4280.88

carolinan:
Tested again with the current master of Gutenberg 9.6 RC1 (5425353), but the result was the same.

ntsekouras:
@carolinan can you reset your database in your testing environment? I'm not sure if something is affected by some template resolution change or something... - needs to be checked. https://developer.wordpress.org/block-editor/packages/packages-env/#5-reset-the-database

scruffian:
I was seeing this at some point with Gutenberg, but on the latest master it works fine.

ntsekouras:
It might probably caused by this: https://github.com/WordPress/gutenberg/pull/27016 (breaking changes). I think it's okay to merge.

carolinan:
If it works for scruffian that is the confirmation I need to merge it :)

PR #148, opened 2020-12-17
TT1 Blocks: Update alignment rules

scruffian:
This is an attempt to simplify our alignment rules. Since blocks are `display: block` they will already take up 100% width. We only need to add margins when a block is _not_ alignfull.

Why wont this work?

### DIFF
--- a/tt1-blocks/style.css
+++ b/tt1-blocks/style.css
@@ -52,11 +52,9 @@ body {
 	margin-right: auto;
 }
 
-.wp-site-blocks .alignwide {
+.wp-site-blocks .wp-block-post-content .alignwide {
 	width: var(--wp--custom--responsive--alignwide-width);
 	max-width: 100%;
-	margin-left: auto;
-	margin-right: auto;
 }
 
 .aligncenter {
--- a/tt1-blocks/style.css
+++ b/tt1-blocks/style.css
@@ -45,13 +45,17 @@ body {
 	padding: 0 var(--wp--custom--spacing--horizontal);
 }
 
-.wp-site-blocks > *:not(.wp-block-post-content):not(.alignfull),
-.wp-site-blocks .wp-block-post-content > *:not(.alignfull) {
-	max-width: var(--wp--custom--responsive--aligndefault-width);
+.wp-site-blocks * {
 	margin-left: auto;
 	margin-right: auto;
 }
 
+.wp-site-blocks > *:not(.wp-block-post-content):not(.alignfull):not(.alignwide),
+.wp-site-blocks .wp-block-post-content > *:not(.alignfull):not(.alignwide) {
+	max-width: var(--wp--custom--responsive--aligndefault-width);
+}
+
+.wp-site-blocks > .alignwide,
 .wp-site-blocks .wp-block-post-content .alignwide {
 	width: var(--wp--custom--responsive--alignwide-width);
 	max-width: 100%;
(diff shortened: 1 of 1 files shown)


kjellr:
I _think_ it will work. But I'm seeing a small bug for wide items in this PR: 

Master Branch|This PR
---|---
Screen Shot 2020-12-17 at 9 43 09 AM|Screen Shot 2020-12-17 at 9 42 46 AM

I think the `alignwide` styles need more specificity to work.

scruffian:
Nice, let me work on it some more then :)

scruffian:
Ok, I made some more changes...

kjellr:
I think we need these rules to apply to blocks both inside and outside of `.wp-block-post-content`. Currently, if something's part of a template part, it doesn't inherit the rules. Note the header and post title here: 

Screen Shot 2020-12-17 at 10 41 46 AM

scruffian:
This is the latest version:
localhost_4759__p=89 preview_id=89 preview_nonce=7ef6cae9d6 preview=true

kjellr:
I'm not seeing the `alignwide` header columns block appearing at the correct width. It looks full-width. Is that just me? 

Screen Shot 2020-12-17 at 11 43 02 AM

scruffian:
No you're right. I've fixed that now.

kjellr:
This doesn't seem to work nicely for nested group blocks, but I don't think it actually has to do with this PR:

Editor|Master Branch|This PR
---|---|---
Screen Shot 2020-12-18 at 9 58 24 AM|Screen Shot 2020-12-18 at 9 58 07 AM|Screen Shot 2020-12-18 at 9 57 51 AM

Mind taking a look to see if this is a quick fix?

scruffian:
I think this means we can simplify the rules more. Try now..

kjellr:
Very cool! This is looking good. I think the last thing is just that wide/full alignments don't look right yet in there: 

Screen Shot 2020-12-18 at 11 03 02 AM

scruffian:
This is happening because the group block inserts an extra wrapper. I think the group block should take responsibility for its own alignment rules since it's doing something strange :)

kjellr:
>  I think the group block should take responsibility for its own alignment rules since it's doing something strange :)

Just like the rest of Gutenberg, It doesn't provide any alignment code at all for the front end. 😞

tt1-blocks/style.css
```diff
@@ -45,36 +45,17 @@ body {
...
-	max-width: calc(100% + (2 * var(--wp--custom--spacing--horizontal)));
-	margin-left: 0;
-	margin-right: 0;
-	box-sizing: content-box;
-}
-
-.wp-site-blocks .wp-block-template-part.alignfull {
-	width: 100%;
-	max-width: 100%;
+.wp-site-blocks *:not(.wp-block-post-content):not(.alignfull):not(.alignwide) {
```
kjellr, line 53:
```suggestion
.wp-site-blocks *:not(.wp-block-post-content):not(.alignfull):not(.alignwide):not([class$="__inner-container"]) {
 {
```

Exclusing inner containers (Group, Cover, etc) from this rule seems to solve the alignment issues noted [here](https://github.com/WordPress/theme-experiments/pull/148#issuecomment-748175575). 

Before|After
---|---
Screen Shot 2020-12-18 at 2 32 31 PM|Screen Shot 2020-12-18 at 2 32 09 PM

scruffian:
well this seems to work pretty nice now, thanks!

kjellr (approved):
Looks good to me! Thanks, @scruffian!

PR #149, opened 2020-12-17
TT1 Blocks: Clean up for Theme Repo

kjellr:
Takes care of a few of the checklist items from https://github.com/WordPress/theme-experiments/issues/140#issuecomment-747372934: 

- Update screenshot
- Update the remaining @subpackage
- Add file header to functions.php

I also added a first draft of a readme.txt. Please suggest any edits!

---

For the screenshot, this was a (very) quick version. @melchoyce I'm curious if you have any ideas. I'm not sure how totally different we want the screenshot, considering it's meant to be the exact same theme visually. 😄 

Original|TT1 Blocks
---|---
screenshot|screenshot


tt1-blocks/readme.txt
```diff
@@ -0,0 +1,59 @@
...
+
+1. If you do not already have <a href="https://wordpress.org/plugins/gutenberg/">the Gutenberg plugin</a> installed and activated, go to Appearance -> Themes in your admin panel and search for 'Gutenberg'. Install and activate the plugin. 
+2. Then, in your admin panel, go to Appearance -> Themes and click the 'Add New' button.
+3. Type in 'TT1 Blocks' in the search form and press the 'Enter' key on your keyboard.
+4. Click on the 'Activate' button to use your new theme right away.
+5. From your admin panel, visit the Site Editor to customize your site.
+
+== Changelog ==
+
+= 1.0 =
```
kjellr, line 26: I'm not sure what to do about the version number here. Should we start below 1.0?

scruffian: 0.1

tt1-blocks/readme.txt
```diff
@@ -0,0 +1,59 @@
+=== TT1 Blocks ===
+Contributors: wordpressdotorg
+Requires at least: 5.6
+Tested up to: 5.6
```
jffng, line 4: [style.css](https://github.com/WordPress/theme-experiments/blob/master/tt1-blocks/style.css#L7-L8) suggests `5.3` for requires and tested versions — should we bump that there also?

kjellr: Ah good call. Yes, I'll update that.

tt1-blocks/readme.txt
kjellr: Works for me. Updated in 582d892.

melchoyce:
How about including the tagline and showing a little more of the image?

screenshot

kjellr:
Works for me. PR is updated.

kjellr:
I just pushed up archive and search results templates: 

Archive|Search Results
---|---
Screen Shot 2020-12-17 at 2 26 55 PM|Screen Shot 2020-12-17 at 2 26 17 PM

Note that there's no title for the archive page yet, because there isn't a block for it yet. 
And the search results page title is static for now. 

I'm not 100% sure about the query block markup there, but it seems to work in my testing.

kjellr:
If those templates look good, I think this will now close #140.

jffng:
I tested the Archive and Search templates against GB master, and looks good to me.

scruffian:
Is there a reason that we need these extra templates? If we need them, maybe we could create another template part which contains the code that is shared across all these templates?

kjellr:
Until the page headers are dynamic, I'm not sure it makes sense to include these additional templates at all — they're basically just the same as the index template (I can easily modify the query there instead of on these two pages, and it will work in all contexts). 

@carolinan, you'd included these two templates [in your list here](https://github.com/WordPress/theme-experiments/issues/140#issuecomment-747372934). What do you think?

tt1-blocks/style.css
```diff
@@ -46,7 +46,8 @@ body {
 }
 
 .wp-site-blocks > *:not(.wp-block-post-content),
-.wp-site-blocks .wp-block-post-content > * {
+.wp-site-blocks .wp-block-post-content > *,
+.wp-block-query-loop li > *:not(.entry-content):not(.alignwide):not(.alignfull) {
```
scruffian, line 50: did this creep in by mistake?

kjellr: No, I just added that. Previously, all content of the query block was showing up as standard width, even when it was wide/full. 😩 I noticed this when I moved the default query into its own template part, and had to do this to fix it.

Here's an example, where the Media & Text block is supposed to be wide width:

Before f4a0131|After I placed the query block in the full-width template part.|Final result, after I added this css.
---|---|---
Screen Shot 2020-12-18 at 9 12 03 AM|Screen Shot 2020-12-18 at 9 11 24 AM|Screen Shot 2020-12-18 at 9 11 09 AM




(I _really_ can't wait until this alignments stuff "just works" 😞).

kjellr:
BTW, I fixed the global query in f4a0131 by using the usual query block markup, and ensuring that it was set to `"inherit":true` (Thanks @ntsekouras for the help). Now the index, archive, and search templates all use the same exact query.

carolinan:
> Until the page headers are dynamic, I'm not sure it makes sense to include these additional templates at all — they're basically just the same as the index template (I can easily modify the query there instead of on these two pages, and it will work in all contexts).
> 
> @carolinan, you'd included these two templates [in your list here](https://github.com/WordPress/theme-experiments/issues/140#issuecomment-747372934). What do you think?

My idea was that they should be included as examples for theme authors to show how the query block can be used
with of the recent changes.
Ideally they should be different, perhaps show excerpt in one and full content in another. And they should be different from index.

kjellr:
I feel like that strays a bit from trying to replicate the original theme. By default, I think Twenty Twenty-One displays only the excerpt everywhere, right? But there's a customizer option to show the full post if desired?

carolinan:
I think it is more important to show what can be done, not to create an exact copy of Twenty Twenty-One, or create a minimal theme (index for everything). There already is a minimal theme.

carolinan:
With this query, pages an posts will have identical search results, no? So it would not match Twenty Twenty-One.

kjellr:
Since the rest of the changes here are solely focused on release, I'll remove the additional templates from this PR and we can consider them separately in https://github.com/WordPress/theme-experiments/issues/152. 

In the meantime, the rest of the changes here should be ready for a review if someone has a moment. The `readme.txt` content could especially use another set of eyes. 

Thanks!

carolinan:
In [functions.php line 216](https://github.com/WordPress/theme-experiments/blob/23d623809a7ffbee60043bebee2547205789305e/tt1-blocks/functions.php#L216) this text domain is incorrect:

`'name'     => esc_html__( 'Red to Purple', 'tt1_blocks_block_editor_script' ),`

tt1-blocks/block-templates/index.html
```diff
@@ -1,6 +1,6 @@
 <!-- wp:template-part {"slug":"header","theme":"tt1-blocks","align":"full", "tagName":"header","className":"site-header"} /-->
 
-<!-- wp:query {"queryId":1,"query":{"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":""}} -->
+<!-- wp:query {"queryId":1,"query":{"perPage":"","pages":"","offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[2227],"sticky":"","inherit":true}} -->
```
carolinan, line 3: Is 2227 a post id? I think it needs to be removed?

tt1-blocks/readme.txt
```diff
@@ -0,0 +1,59 @@
...
+the Free Software Foundation, either version 2 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU General Public License for more details.
+
+
+TT1 Blocks is derived from Twenty Twenty-One, (C) 2020 WordPress.org
```
carolinan, line 47:
```suggestion
TT1 Blocks is derived from Twenty Twenty-One, (C) 2020 WordPress.org. License: GPL v2 or later.
```

carolinan:
Unless we want to add links to information about the outreach program, I think the readme is fine.

If the is any code borrowed from for example Bosco, Gutenberg starter theme Q or emptytheme then those themes and theme developers should be credited too.

carolinan:
I don't think these two are needed since they are also set in the json file?
```
add_theme_support( 'custom-line-height' );
add_theme_support( 'custom-spacing' );
```

kjellr:
> In functions.php line 216 this text domain is incorrect:

Thanks, should be [fixed](https://github.com/WordPress/theme-experiments/pull/149/commits/d0cedb8c3029c9d06433352ed2198f5c85766929) now. 

> If there is any code borrowed from for example Bosco, Gutenberg starter theme, Q or emptytheme then those themes and theme developers should be credited too.

I don't think there is anymore. The alignments [came from here originally](https://github.com/WordPress/gutenberg/issues/25160#issuecomment-691076081), but they've been rewritten by @scruffian recently. The rest of the base of the theme was just built from scratch or copy/pasted from the original Twenty Twenty-One. 

> I don't think these two are needed since they are also set in the json file?

Good call. 👍 

This should be ready for another review!

carolinan:
There are functions in the removed header.php and footer.php files that are needed to pass theme check for the upload to the theme directory.

carolinan:
@Otto42 since this theme will be added to the theme directory from the wordpressorg account does it still need to pass theme check?

carolinan:
The post comments block uses [comments_template](https://github.com/WordPress/gutenberg/blob/master/packages/block-library/src/post-comments/index.php#L31) so the theme still needs a comments.php file:

`Deprecated: Theme without comments.php is deprecated since version 3.0.0 with no alternative available. Please include a comments.php template in your theme. in /var/www/html/wp-includes/functions.php on line 5059`

kjellr:
> There are functions in the removed header.php and footer.php files that are needed to pass theme check for the upload to the theme directory.

It's possible I added those intentionally and forgot why over the holidays. 😄 

In any case, even with those files, the theme fails ThemeCheck with a ton of errors: 

Screen Shot 2021-01-05 at 8 32 59 AM

So if it does need to pass ThemeCheck, we need to add many more (unused) selectors and functions. If possible, it would be a much better solution to fix ThemeCheck so it doesn't require all this stuff for block based themes.

carolinan:
Without finalized requirements we can't update Theme Check -we won't know what it needs to check for. I estimate that to take at least 3 months.

carolinan:
Ah, no that is not how the theme upload works, you have have to pass Theme Check **once**.
Once the theme data is saved (it is a custom post type) and the files are uploaded, the moderators need to log in and enable a "Special Case Theme" setting in the admin area. This means that theme is not checked by Theme Check on the next upload.
**After that, the files can be removed from the theme.**

That is why I wanted to ask Otto because I believe he mentioned that there is an exception for default themes. This is not a default or bundled theme but it will be uploaded from the same account.

kjellr:
Ok cool. I think the plan was to upload a generic theme with the TT1 Blocks name to start with (a version that passes ThemeCheck), and then once a moderator enables that special case, we can upload the actual version. That should work, right?

carolinan:
Yes.

kjellr:
Cool. If that's the case, I think we can leave out those `header.php` and `footer.php` files. If folks are ok with the rest of what's in this PR, we should be able to get the submission process moving soon.

carolinan:
I was hoping to be able to merge this now but when I view an archive or search result I see the following:

`Warning: A non-numeric value encountered in /var/www/html/wp-content/plugins/gutenberg/build/block-library/blocks/query-loop.php on line 45`

Line 45 is the per page setting:
https://github.com/WordPress/gutenberg/blob/master/packages/block-library/src/query-loop/index.php#L45

In index.html, the per page setting is `"query":{"perPage":"",`

I am not sure if it has been discussed already but the default value for $posts_per_page is 10 and I think this should reflect that.

tt1-blocks/block-templates/index.html
```diff
@@ -1,6 +1,9 @@
 <!-- wp:template-part {"slug":"header","theme":"tt1-blocks","align":"full", "tagName":"header","className":"site-header"} /-->
 
-<!-- wp:query {"queryId":1,"query":{"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":""}} -->
+<!-- wp:query {"queryId":1,"query":{"perPage":"10","pages":"","offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":"","inherit":true}} -->
+
+<!-- wp:query-loop -->
```
carolinan, line 5:
```suggestion

```

carolinan:
Has the lack of query pagination been discussed? Is not including pagination something you as feel strongly about as the comments?

The pagination does not have a CSS class on the front so we can not add any styles for it until the block is updated. But the navigating between the archive pages technically works.

scruffian:
There are a couple of open issues for pagination:
https://github.com/WordPress/gutenberg/issues/26557
https://github.com/WordPress/gutenberg/issues/24762

carolinan:
Yes but they do not address why the block is not being used in the theme.

scruffian:
I created an issue to add the block to this theme: https://github.com/WordPress/theme-experiments/issues/158

kjellr:
> Yes but they do not address why the block is not being used in the theme.

There's a simple answer actually: I tried adding it a while back, but it didn't seem to work. 😂 

Let's look at sorting that out in another PR — maybe it's working now.

Issue #146, opened 2020-12-16
Empty Theme: Templates reference wrong theme slug for header.html


justintadlock:
`index.html` and `singular.html` reference the `emptytheme` slug:

```
<!-- wp:template-part {"slug":"header","theme":"emptytheme","tagName":"header"} /-->
```

This is not updated when the theme is generated through the command line tool to match the generated theme's slug.

kjellr:
@justintadlock I think that was fixed this morning in #143. Can you confirm?

Thanks!

justintadlock:
Looks like I pulled the repo just before this change went in.

But, there's still an issue with the slug.  Using `Test 1` (a two-word theme name), the slug has an underscore instead of hyphen:

```
<!-- wp:template-part {"slug":"header","theme":"test_1","tagName":"header"} /-->
```

kjellr:
> Using Test 1 (a two-word theme name), the slug has an underscore instead of hyphen

Good catch, thanks! @MaggieCabrera or @pbking is that something you can take a look at?

MaggieCabrera:
> > Using Test 1 (a two-word theme name), the slug has an underscore instead of hyphen
> 
> Good catch, thanks! @MaggieCabrera or @pbking is that something you can take a look at?

Yep, I will have a look!

PR #150, opened 2020-12-17
Empty Theme: Changed slug on html templates

MaggieCabrera:
This PR changes the slug used on html templates to be the one using dashes instead of underscores.

Closes https://github.com/WordPress/theme-experiments/issues/146

### DIFF
--- a/new-empty-theme.php
+++ b/new-empty-theme.php
@@ -58,7 +58,7 @@ class Generate_Theme {
 			
 			mkdir($dir);
 			$prototype_dir = $this->old_themeslug . '/';
-			$exclude_files = array( '.travis.yml', 'codesniffer.ruleset.xml', '.jscsrc', '.jshintignore', 'CONTRIBUTING.md', '.git', '.svn', '.DS_Store', '.gitignore', '.', '..' );
+			$exclude_files = array( '.travis.yml', 'codesniffer.ruleset.xml', '.jscsrc', '.jshintignore', 'README.md', 'CONTRIBUTING.md', '.git', '.svn', '.DS_Store', '.gitignore', '.', '..' );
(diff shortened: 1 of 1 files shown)


kjellr:
This is working well! One thing I noticed though: The enqueue handles are written out using underscores, but usually those use dashes. Would it be possible to change that too? All the other strings in that file would still use underscores, so I know it may be tougher. 

Screen Shot 2020-12-17 at 11 49 25 AM

kjellr:
Also, this is not really related, but it might make sense to not copy over the `README.md` file. Since that's not required for the new theme, and isn't modified by the script anyway.

MaggieCabrera:
> This is working well! One thing I noticed though: The enqueue handles are written out using underscores, but usually those use dashes. Would it be possible to change that too? All the other strings in that file would still use underscores, so I know it may be tougher.
> 
> Screen Shot 2020-12-17 at 11 49 25 AM

I can totally do that, I thought the underscores were the default for the function names but it's easy to change them really.

kjellr:
> I can totally do that, I thought the underscores were the default for the function names but it's easy to change them really.

It is the default for the function names — just not for the enqueue function handles ('two_words-alignments-style` would usually be `two-words-alignments-style` for example).

MaggieCabrera:
> > I can totally do that, I thought the underscores were the default for the function names but it's easy to change them really.
> 
> It is the default for the function names — just not for the enqueue function handles (`two_words-alignments-style` would usually be `two-words-alignments-style` for example).

Oh yeah, I get what you mean! I'll try that

MaggieCabrera:
This is ready for review. Regex are not my strong suit so please review away :D

kjellr (approved):
This works in my testing! Thanks, @MaggieCabrera!

PR #155, opened 2020-12-27
Add theme phoenix

imranhsayed:
Adds Theme Phoenix - an experimental block based theme.

### Features and Description
1. It uses the Navigation block in the header.
2. Front page block template contains
- Hero section
- Latest posts using Query Block
- Sidebar containing Gallery, Social Links, Categories and Search Blocks.
3. Single page, Single post and 404 block templates.

All the sections are editable from WordPress Dashboard > Site Editor

[Live Demo Site](https://phoenix.codeytek.com/)

screencapture-phoenix-codeytek-2020-12-27-13_56_01-1

image


phoenix/block-templates/404.html
```diff
@@ -0,0 +1,23 @@
+<!-- wp:template-part {"slug":"header","theme":"phoenix","tagName":"header"} /-->
```
pbking, line 1: To satisfy [recent changes in Gutenberg](https://github.com/WordPress/gutenberg/pull/28088) (as of 9.9) the "theme" attribute should be removed from the template-part blocks.

pbking:
It's interesting to see a FSE theme that is NOT using the (experimental-)theme.json.  It absolutely works in wp-admin and the views, however without it the Full Site Editor itself doesn't load.  I just copied the [simple theme.json from emptytheme](https://github.com/WordPress/theme-experiments/blob/master/emptytheme/experimental-theme.json) and that was enough to make things work, however you may want to adjust things further before bringing this in.

imranhsayed:
Thank you for your feedback @pbking .
I have implemented the feedback and added the `(experimental)-theme.json` with settings as per the theme requirements.
The global styles work great with Gutenberg v9.8.4 . However the v9.9 has some issues with the global styles application.

pbking:
> The global styles work great with Gutenberg v9.8.4 . However the v9.9 has some issues with the global styles application.

There were breaking changes regarding the JSON structure.  You can read up more about that here: https://make.wordpress.org/themes/2021/02/04/gutenberg-9-9-new-json-structure-for-theme-json-files/

I should have mentioned that any only just realized that the version in empty themes is the the old structure.

imranhsayed:
Thank you for the update @pbking . I have made the necessary changes as suggested.

imranhsayed:
Hi @pbking,
Please let me know if there are any other changes required 😊

PR #156, opened 2020-12-29
Probd blocks

AbmSourav:
proBD Blocks is a block-based version of the proBD theme.


pbking:
There seem to be some PHP errors from setup.php where the color palettes are being added.  This seemed to bring the whole site editor down so I couldn't evaluate much further.

As an alternative to adding those color palettes via PHP I suggest instead including them exclusively in the experimental-theme.json file instead.  I don't believe there would be a reason in your theme to have them in both locations.  As a test I simply removed 

```
 31     $this->editor_font_size();
 32     
 33     $this->editor_color_palette();
 34     
 35     $this->editor_gradient_palette();
```
from [setup.php](https://github.com/WordPress/theme-experiments/pull/156/files?file-filters%5B%5D=.php#diff-dadb00b54b5cd39d78153f5d5ac334d7d638e298192c86ca9401d8a26534dfe8R31) and things seemed to run OK.

carolinan:
Hi, do you plan to update this to the newest formats?

Issue #158, opened 2021-01-06
TT1 Blocks: Add the query pagination block


scruffian:
We should add the query pagination block to archive pages so users can navigate between them.

From: https://github.com/WordPress/theme-experiments/pull/149#issuecomment-755165588

PR #160, opened 2021-01-07
TT1 Blocks -Add query pagination

carolinan:
Fixes https://github.com/WordPress/theme-experiments/issues/158

Adds the query pagination block to index.php.
Updates the query page setting to 100 so that there are pages of results to show 
(It is not possible to select for example  -1 to display all existing results).


kjellr (approved):
Works in my testing. Thank you!

Screen Shot 2021-01-07 at 9 38 27 AM

PR #162, opened 2021-01-07
Fix #159

### LINKED ISSUE #159: Cover Block render error when empty-theme is active
For some reason, the cover block will fail to render when the empty theme is active. Activating any other of the experimental themes solves the issue.

carolinan:
Fix #159

Add support for custom units to emptytheme.


kjellr (approved):
Thanks for the fix, @carolinan!

PR #168, opened 2021-01-14
TT1 Blocks: Version bump + update tags

kjellr:
As per @carolinan's notes [on slack](https://wordpress.slack.com/archives/C02RP4VMP/p1610620845005800): 

> But there are actually other tags that are incorrect as well. footer-widgets, translation-ready are not available in FSE.
> The theme needs to have the full-site-editing tag because it is used to filter the themes in the theme directory.

This PR removes those two tags, and adds the full-site editing tag. 

It also bumps the version number so we can submit this to the theme repo. Rather than do a whole point release, I've changed this to 0.4.1... That way it'll be a while before we hit 1.0. 😄

### DIFF
--- a/tt1-blocks/readme.txt
+++ b/tt1-blocks/readme.txt
@@ -23,7 +23,7 @@ This theme is beta software, and is not meant for use on a production site. Bug 
 
 == Changelog ==
 
-= 0.4 =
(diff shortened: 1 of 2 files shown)


scruffian (approved):
LGTM

Issue #170, opened 2021-01-14
TT1 Blocks: Fix alignment on Cover Block


beafialho:
While testing TT1 Blocks, I've encountered a bug on the Cover Block when choosing the Full Width option. Whether it's media or a color background, it doesn't stretch to the edge of the screen.

<img width="1536" alt="Captura de ecrã 2021-01-14, às 16 25 40" src="https://user-images.githubusercontent.com/65220155/104619057-5b758680-5685-11eb-8953-ac79e9745685.png">

Jastos:
This will also happen on regular image blocks. 

`.wp-site-blocks *:not(.wp-block-post-content):not(.alignfull):not(.alignwide):not([class$="__inner-container"]) {
	max-width: var(--wp--custom--responsive--aligndefault-width);
}`

Does not allow the img tag to be full width since it is a child of the .alignfull on the figure element. A quick fix would be to add another :not, but this probably isn't the best solution.

`.wp-site-blocks *:not(.wp-block-post-content):not(.alignfull):not(.alignwide):not([class$="__inner-container"]):not(figure img) {
	max-width: var(--wp--custom--responsive--aligndefault-width);
}`

PR #174, opened 2021-01-22
TT1 Blocks: Fix frontend full alignment styles

kjellr:
Fixes https://github.com/WordPress/theme-experiments/issues/170

The `max-width` limit here should not override the width of images here, since they should generally be the width of their `figure` or cover containers. 

Separately, we weren't specifying a `max-width` for `alignfull`, which means that they were maxing out at `100%` by default and not taking into account the left/right padding on the container. 

## Screenshots

Before

Screen Shot 2021-01-22 at 9 08 51 AM

After

Screen Shot 2021-01-22 at 9 23 23 AM

### DIFF
--- a/tt1-blocks/style.css
+++ b/tt1-blocks/style.css
@@ -50,7 +50,7 @@ body {
 	margin-right: auto;
 }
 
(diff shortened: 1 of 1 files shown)


scruffian (approved):
LGTM

Issue #167, opened 2021-01-13
TT1 Blocks: Fix or remove front page template


fklein-lu:
The TT1 has a `front-page.html` template that only shows the post content.

There are two main issues with this:

1. The original theme does not seem to have a front page template.
2. But front page templates need to respect the reading settings. This means that if the homepage is set to display posts, a list of posts should be displayed. So there most be a loop and a blog display here.

carolinan:
This need to be solved with the query block(?) up streams, and the theme can be adjusted afterwards.

fklein-lu:
No, the template needs to be fixed. Because there's no Query Block in there.

Add a Query Block to the template, have it inherit the Main Query, and it all will work as expected.

Or better yet remove the template, since the original Twenty Twenty-One theme doesn't have a front page template either, and TT1 is supposed to be a block version of that theme.

PR #179, opened 2021-01-26
TT1 Blocks: Delete front-page.html

carolinan:
Solves https://github.com/WordPress/theme-experiments/issues/167

### DIFF
--- a/tt1-blocks/block-templates/front-page.html
+++ b/tt1-blocks/block-templates/front-page.html
@@ -1,5 +1,0 @@
-<!-- wp:template-part {"slug":"header","theme":"tt1-blocks","align":"full", "tagName":"header","className":"site-header"} /-->
-
-<!-- wp:post-content /-->
-
-<!-- wp:template-part {"slug":"footer","theme":"tt1-blocks","align":"full","tagName":"footer","className":"site-footer"} /-->


kjellr:
I think we _do_ actually need a separate template here — since Twenty Twenty-One hides the homepage title on the homepage automatically. It should probably have a query block, but with no page title.

carolinan:
If we did that, the list of blog posts would also not have titles.

We do not have traditional PHP conditionals, but we do have CSS. We **can** work around it, but a long term solution
needs to be figured out for the editor.

.home:not(.blog) .wp-block-post-title,
.home:not(.blog) .wp-block-post-comments,
.home:not(.blog) .wp-block-post-comments-form {
	display: none;
}

This would leave the .wp-block-separator above the content.
Without a wrapper for the "post header" we can not just target the first separator of the page.

That the `main` element is missing is also a big problem.

kjellr:
Hmm. That's a good point. Perhaps for now we do remove the `front-page.html` template, and discuss how to support the separate "single page w/no title" template in a separate issue. cc @scruffian for thoughts too.

scruffian:
Yeah we will definitely need to fix this in Gutenberg. I created an issue here: https://github.com/WordPress/gutenberg/issues/28564

Happy to remove the template for now, but I'd rather not fix it with CSS.

kjellr (approved):
Sounds good. This should be good to merge in the meantime!

mtias:
Agreed front-page doesn't make much sense. The relationship between front-page, home, page template, and the static homepage setting are all a bit confusing in WordPress. Would it make sense to have a `page-home` template for the case of no title?

I think more generally we need make the "hide a title" something much more flexible across the board, for pages and posts alike — maybe a part of the post-title block setting a body-class or post-class.

scruffian:
That's what I did here: https://github.com/WordPress/theme-experiments/pull/184/files

PR #183, opened 2021-01-28
TT1 Blocks: allow custom overrides for base and heading fonts

mattwiebe:
This will allow base fonts to be overridden by supplying the `--font-base` CSS custom property on `:root`
and headings to alternately be overridden by `--font-headings`, falling back to the base font if not supplied.

The current goal is to enable customization in the Gutenboarding flow on WordPress.com, and will fix the issue
of it not working detailed at https://github.com/Automattic/wp-calypso/issues/48973

Testing should simply be that nothing is changed by default. You can use your inspector to add custom properties. If you add `--font-base` to the `html` element, you should see all of the fonts change to render with that value, and adding `--font-headings` should switch all headings to that value. 

Eg, with `--font-base` set to `Georgia`:
image

`--font-base: Georgia; --font-headings: Verdana;` 👍 
image

### DIFF
--- a/tt1-blocks/assets/css/style-shared.css
+++ b/tt1-blocks/assets/css/style-shared.css
@@ -25,8 +25,12 @@ a:hover {
 }
 
 /*
- * Gutenberg remotes this underline, but Twenty Twenty-One uses it. 
+ * Gutenberg remotes this underline, but Twenty Twenty-One uses it.
  */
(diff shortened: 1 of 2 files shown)


kjellr:
Closing as per the notes in https://github.com/Automattic/wp-calypso/issues/48973#issuecomment-769687431.

PR #184, opened 2021-01-29
TT1 Blocks: Add a static homepage template

scruffian:
This adds a specific template for the home page, without the page title, so that when its set to be the front page of the site we can hide the title without resorting to CSS.

Follow up to #179


carolinan:
Hi
For the template to be selectable in the page attributes in the document sidebar, it also needs to be added to the .json file:
```
"pageTemplates": {
		"page-home": {
			"title": "Page without title" (or what ever we want here) -
		}
	}
```
On the root level.

scruffian:
Ah I didn't realise that! Thanks. I added a commit for it.

Issue #180, opened 2021-01-28
TT1 Blocks: Main element is missing


carolinan:
The header template part and footer template part correctly uses `<header>` and `<footer>` elements
But in our wishes to keep the theme simple, the `<main>` element was not used to wrap the main content.

This is an oversight that has negative impacts on accessibility and it needs to be resolved.
More info: https://dequeuniversity.com/rules/axe/4.1/landmark-one-main

carolinan:
Since the query is not a wrapping element -it has no element,
the options are to wrap the queries in a group block or use template parts.

Template parts are practical for repeated content, I suggest that the simplicity of this theme makes use of group blocks more appropriate.

kjellr:
> I suggest that the simplicity of this theme makes use of group blocks more appropriate.

I agree, that makes sense!

carolinan:
I thought it had been discussed and decided against already.

kjellr:
Ah perhaps — It's totally possible that I missed it. 😄

mtias:
> Template parts are practical for repeated content, I suggest that the simplicity of this theme makes use of group blocks more appropriate.

Agreed. We have discussed a _main_ or _content_ semantic template part, but it's a bit of overhead for something that seems simpler. Starting with group + tag seems great to me. It's always possible to convert to a template part later if needed.

PR #190, opened 2021-02-01
TT1 Blocks: Add main element

carolinan:
Fixes https://github.com/WordPress/theme-experiments/issues/180

Adds a full width group block with `"tagName":"main"` around the content.


kjellr (approved):
Works in my testing!

PR #196, opened 2021-02-04
TT1 Blocks: allow custom font overrides

mattwiebe:
Hi, I'm back again! This is nearly/exactly identical to #183 and will unblock broader testing of this theme on WordPress.com, where we allow font customizations. 

This will allow base fonts to be overridden by supplying the `--font-base` CSS custom property on `:root`
and headings to alternately be overridden by `--font-headings`, falling back to the base font if not supplied.

The current goal is to enable customization in the Gutenboarding flow on WordPress.com, and will fix the issue
of it not working detailed at https://github.com/Automattic/wp-calypso/issues/48973

Testing should simply be that nothing is changed by default. You can use your inspector to add custom properties. If you add `--font-base` to the `html` element, you should see all of the fonts change to render with that value (allowing a single font for the whole theme, as is currently the case), and adding `--font-headings` should switch all headings to that value. 

Eg, with `--font-base` set to `Georgia`:
image

`--font-base: Georgia; --font-headings: Verdana;` 👍 
image

### DIFF
--- a/tt1-blocks/assets/css/style-shared.css
+++ b/tt1-blocks/assets/css/style-shared.css
@@ -8,6 +8,10 @@ body {
 	-webkit-font-smoothing: antialiased;
 }
 
+h1, h2, h3, h4, h5, h6 {
+	font-family: var(--wp--custom--font-secondary), var(--wp--custom--font-primary);
+}
+
 /*
  * text-underline-offset doesn't work in Chrome at all 👎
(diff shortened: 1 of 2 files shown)


carolinan:
It looks like this has been solved on the calypso side, can this be closed?

kjellr:
Yes, it can be. Thanks!

PR #199, opened 2021-02-07
Update Armando

carolinan:


armando/block-template-parts/content-single.html
```diff
@@ -1,6 +1,6 @@
 <!-- wp:post-title {"className":"default-width"} /-->
 
-<!-- wp:template-part {"slug":"meta","theme":"armando","backgroundColor":"silver","className":"post-meta default-width is-style-armando-box-shadow"} /-->
+<!-- wp:template-part {"slug":"meta","backgroundColor":"silver","className":"post-meta default-width is-style-armando-box-shadow"} /-->
```
carolinan, line 3: It seems that template parts included in template parts are not found, unless the theme attribute is included.
Can anyone confirm?
(Guteberg 9.9.0, no subfolder)

ndiego: Yes, I can confirm. I just finished removing all my theme attributes and just discovered this!

carolinan: I am creating an issue right now

PR #202, opened 2021-02-08
Rename pageTemplates to customTemplates

nosolosw:
Updates the key used to declared custom templates as per https://github.com/WordPress/gutenberg/pull/28830

### DIFF
--- a/tt1-blocks/experimental-theme.json
+++ b/tt1-blocks/experimental-theme.json
@@ -305,7 +305,7 @@
 			}
 		}
 	},
-	"pageTemplates": {
+	"customTemplates": {
 		"page-home": {
 			"title": "Page without title"
 		}


nosolosw:
cc @kjellr @carolinan not sure what's the status of updating TT1-blocks in the repo with the latest changes to the theme.json structure, but there's another PR that updates a name key so it could be bundled with that update if hasn't been done yet.

kjellr:
> cc @kjellr @carolinan not sure what's the status of updating TT1-blocks in the repo with the latest changes to the theme.json structure,

The structure should be up to date now, so this would be a separate change. I suggest we merge for now, and we can try to bundle this change with a theme version that's released alongside Gutenberg v10.0.

PR #204, opened 2021-02-10
TT1 Blocks: Patterns: Fix spelling: Roses Trémières

mcsf:
Use correct accented spelling for the hollyhock flower.

See: https://en.wiktionary.org/wiki/rose_trémière


kjellr:
cc @melchoyce — I think this needs to be fixed in core too?

melchoyce:
Yeah, it should get fixed in both locations.

kjellr:
Thanks! I've filed [a core issue here](https://core.trac.wordpress.org/ticket/52500), and I'll merge this in for now. 👍

PR #205, opened 2021-02-12
TT1-Blocks: Fix full width alignment for page-home post content

jeyip:
## Description
On initial creation, the `page-home` template appears to have content that doesn't respect `wide width` or `full width` alignment.

Screenshot 2021-02-01 at 23 31 04

## Testing
- Spin up the site editor locally
- Activate the tt1-blocks theme
- Navigate to the site editor
- Select the `page-home` template to display
- Find the post content block and click on it
- Verify that the alignment defaults to "full"

### DIFF
--- a/tt1-blocks/block-templates/index.html
+++ b/tt1-blocks/block-templates/index.html
@@ -9,7 +9,7 @@
 		<!-- wp:query-loop -->
 		<!-- wp:post-title {"isLink":true} /-->
 
-		<!-- wp:post-content /-->
+		<!-- wp:post-content {"align":"full"} /-->
 
 		<!-- wp:spacer {"height":70} -->
 		<div style="height:70px" aria-hidden="true" class="wp-block-spacer"></div>
--- a/tt1-blocks/block-templates/page-home.html
+++ b/tt1-blocks/block-templates/page-home.html
@@ -3,7 +3,7 @@
 <!-- wp:group {"tagName":"main", "align":"full"} -->
 <main class="wp-block-group alignfull">
 	<div class="wp-block-group__inner-container">
-		<!-- wp:post-content /-->
+		<!-- wp:post-content {"align":"full"} /-->
 	</div>
 </main>
 <!-- /wp:group -->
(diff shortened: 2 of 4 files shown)


kjellr:
Is this an issue on TT1 Blocks' other page templates too? They don't seem to have this attribute on the post content.

vindl:
> Is this an issue on TT1 Blocks' other page templates too? They don't seem to have this attribute on the post content.

@kjellr there is more info about it in this comment https://github.com/Automattic/wp-calypso/issues/49573#issuecomment-777923802. It doesn't seem to be happening with other page templates.

kjellr:
Very weird! This PR appears to work so I'm ok merging, but I'd like a quick gut check from @scruffian just to be sure.

scruffian (approved):
LGTM

jeyip:
@kjellr @scruffian 

Would we be comfortable adding an `align: full` attribute to other post-content blocks in tt1-blocks? Jacopo brought up some great points [here](https://github.com/Automattic/wp-calypso/issues/49573#issuecomment-778292664):

> Considering the Post Content role, I'm actually inclined to default it to full width, freeing its content up to align as they see fit. I might be overlooking something, but this is basically [the same we already do with the Group block](https://github.com/WordPress/theme-experiments/blob/5a02927f9198f903211aba44c009fb2ba793472d/tt1-blocks/block-templates/page-home.html#L3) wrapping the page content inside a `<main>` tag:
```html
<!-- wp:group {"tagName":"main", "align":"full"} -->
<main class="wp-block-group alignfull">
		<!-- wp:post-content /-->
	</div>
</main>
<!-- /wp:group -->
```

> Although, I'm also reluctant to assign a default alignment to a block, as I think this is more of a theme responsibility.
We already `align: full` the Group block in all templates, I guess we should just do the same for Post Content

vindl (approved):

kjellr (approved):
> Would we be comfortable adding an align: full attribute to other post-content blocks in tt1-blocks?

Yeah, I think that makes sense, @jeyip. This works fine in my testing, so let's merge and go from there. 👍

PR #210, opened 2021-02-16
Remove Color Palette & Font Sizes from PHP

pbking:
Since the Editor Color Palette and Editor Font Sizes are expressed in
theme.json there isn't a need to ALSO express that configuration via
PHP.

This change removes the PHP configuration relying exclusively on JSON
configuration.

#### Testing

To test load the Gutenberg Block Editor and select a Button (or another block that allows for color customizations that include gradient) and note that the palette options available before this change remains unchanged.

image

Note the same for available Font Sizes:
image


kjellr (approved):
LGTM! Thanks

Issue #126, opened 2020-12-09
TT1 Blocks: Pullquote Block: Migrate CSS to theme.json


scruffian:
The code is [here](https://github.com/WordPress/twentytwentyone/tree/trunk/assets/sass/05-blocks/pullquote).

From #82

PR #225, opened 2021-02-25
TT1 Blocks: Migrate Pullquote block styles

danieldudzic:
Part of #82
Closes: #126

Dependencies: 
* Adding `typography` support for Pullquotes
* Adding `border-width` support for Pullquotes (there's an Issue about adding support for the Table block: https://github.com/WordPress/gutenberg/issues/27651)
* Adding `padding` support for Pullquotes  (there are already conversations about adding support for other blocks: Code: https://github.com/WordPress/gutenberg/pull/27582 | Verse: https://github.com/WordPress/gutenberg/pull/27579)

PR: https://github.com/WordPress/gutenberg/pull/29361


kjellr:
I refreshed this PR today. I removed the font size stuff since it still hasn't landed in Gutenberg, and was making the citation line really huge. Here's the current state: 

TT1 Blocks|Twenty Twenty-One
---|---
Screen Shot 2021-08-19 at 11 51 18 AM|Screen Shot 2021-08-19 at 11 51 13 AM

As you can see, we're missing the ability to set the border to appear on just the top and bottom. I've opened a Gutenberg issue here to discuss that: https://github.com/WordPress/gutenberg/issues/34179

pbking:
> we're missing the ability to set the border to appear on just the top and bottom.

If we set the value of `border.style` to `"solid none"` that appropriately renders the borders like TwentyTwenyOne's:



And while the user wouldn't be able to reproduce the effect with the editor they can still switch to any of the other options just fine.

image

I agree that the work in Gutenberg is still needed but I don't believe '"solid none"' is wrong.  What do you think?

pbking:
Add in a dash of typography styles and it's not a far throw from the original:
image

kjellr (approved):
That works for me! Thanks, @pbking!

PR #228, opened 2021-03-02
Update index.php

bph:
It explains, why there is nothing to see.
~DocumentsCleanShot 2021-03-02 at 12 39 43

### DIFF
--- a/tt1-blocks/index.php
+++ b/tt1-blocks/index.php
@@ -5,6 +5,6 @@
  */
 
 if ( current_user_can( 'activate_plugins' ) ) {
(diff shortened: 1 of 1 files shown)


kjellr:
This makes sense. I wonder if this should be printed as a translatable string, but I might be overthinking it. 😄 I'll let someone else weigh in on that too.

carolinan:
Yes all theme text strings need to be translation ready :)
Example https://github.com/WordPress/theme-experiments/blob/master/armando/index.php

bph:
@carolinan and @kjellr Thank :-)

kjellr (approved):
Looks good to me. Thank you @bph!

PR #229, opened 2021-03-04
TT1 Blocks: Remove the Dark Gray social links block style

### LINKED ISSUE #176: TT1 Blocks: Remove the Dark gray social link block style
With the added [color options to the social links block](https://github.com/WordPress/gutenberg/pull/28084
), the block style is no longer needed, it is subpar to the new color settings.

And when the block style is selected in the block setting sidebar, the color options has no effect in the editor, only on the front,
this may cause confusion.

A question arises on how one can "migrate" existing block styles, for users who has already selected this style.
Especially if we consider making the same change to Twenty Twenty-One, once this option is added to WordPress core.

danieldudzic:
TT1 Blocks: Remove the Dark Gray social links block style in favor of color options. Closes #176

### DIFF
--- a/tt1-blocks/inc/block-styles.php
+++ b/tt1-blocks/inc/block-styles.php
@@ -98,15 +98,6 @@ if ( function_exists( 'register_block_style' ) ) {
 				'label' => esc_html__( 'Thick', 'tt1-blocks' ),
 			)
 		);
-
-		// Social icons: Dark gray color.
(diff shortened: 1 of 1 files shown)


carolinan:
Hi
I think we can safely remove the CSS and the CSS classes from the footer template and  blockpatterns as well.
We don't need to consider backwards compatibility for the 100 active installs.

PR #232, opened 2021-03-18
TT1 Blocks: Update custom templates and template parts in the config

gziolo:
Related to changes proposed in Gutenberg in https://github.com/WordPress/gutenberg/pull/29828.

There is going to be a bit different format for `customTemplates` and `templateParts`. The rationale behind it is that we want to align with other lists like font sizes, color palettes, or gradients. It should make it easier to write those configurations but it also will simplify the logic that handles translations.


carolinan:
Works well for me.

PR #211, opened 2021-02-16
Added (numerical) Font Weight Value to Headings

pbking:
As a replacement for #135, this change adds font weights to the headers via theme.json via numerical values rather than CSS variables.

Closes #97 

This is necessary due to [this Gutenberg change](https://github.com/WordPress/gutenberg/pull/27555).

This change (as was the original) is still dependent on this Gutenberg change allowing the Heading block to adjust the font weight: https://github.com/WordPress/gutenberg/pull/27639

PR #239, opened 2021-03-23
TT1 Blocks Font Weight Value to Headings

carolinan:
Alternative to #135 and #211, using a custom variable for the font weight for the title blocks.

Since https://github.com/WordPress/gutenberg/pull/29941, neither this PR or #211 relies on https://github.com/WordPress/gutenberg/pull/27639

### DIFF
--- a/tt1-blocks/experimental-theme.json
+++ b/tt1-blocks/experimental-theme.json
@@ -190,7 +190,7 @@
 					"vertical": "30px"
 				},
 				"font-weight":{
-					"title": "300"
+					"thin": "300"
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


tt1-blocks/experimental-theme.json
```diff
@@ -188,6 +188,9 @@
 					"unit": "20px",
 					"horizontal": "25px",
 					"vertical": "30px"
+				},
+				"font-weight":{
+					"title": "300"
```
kjellr, line 193: I prefer this approach, but I'd recommend we call this `light` instead of `title`, just in case we use it outside of just titles.

carolinan: meh where did I get "thin" from -switching again, to "light".

carolinan:
I double checked with Twenty Twenty-One:

h1-h3: normal
h4-h6: strong

site title: normal

h1.entry-title (single view post/page title): 300

kjellr (approved):
This is working well for me. Thanks!

PR #253, opened 2021-04-29
Added theme.json styles for button block to (better) align with TwentyTwentyOne button styles

pbking:
Target:
image

Before:
image

After:
image

Editor:
image

To test add a button to a page. 
Add another button to a page and set the block style to "outline".
Observe that the styles are (closer to) TwentyTwentyOne button styles.

### DIFF
--- a/tt1-blocks/experimental-theme.json
+++ b/tt1-blocks/experimental-theme.json
@@ -266,7 +266,7 @@
 					"text": "var(--wp--preset--color--green)"
 				},
 				"typography": {
-					"fontSize": "22.5px",
+					"fontSize": "var(--wp--preset--font-size--normal)",
 					"fontWeight": 500
 				}
 			},
--- a/tt1-blocks/experimental-theme.json
+++ b/tt1-blocks/experimental-theme.json
@@ -257,6 +257,19 @@
 					"lineHeight": 1.4
 				}
 			},
+			"core/button": {
+				"border": {
+					"radius": "0px"
+				},
+				"color": {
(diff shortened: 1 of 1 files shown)


tt1-blocks/experimental-theme.json
```diff
@@ -208,6 +208,19 @@
...
+		"core/button": {
+			"border": {
+				"radius": "0px"
+			},
+			"color": {
+				"background": "var(--wp--preset--color--gray)",
+				"text": "var(--wp--preset--color--green)"
+			},
+			"typography": {
+				"fontSize": "22.5px",
```
scruffian, line 220: Should we use one of the preset font sizes? Or add this to them?

pbking: I had that debate in my head and went back and forth.

I finally landed on the above to achieve design parity with the original and didn't add it as a preset font size, again to achieve feature parity with TwentyTwentyOne (as it [doesn't provide that one)](https://github.com/WordPress/twentytwentyone/blob/trunk/functions.php#L146).

So my vote is no to both.

scruffian: What about the same question for font-weight?

pbking: That one I'm impartial about.  Personally I don't really see much value in making it a variable but I can't argue that it shouldn't be.

pbking: 👍 ✅

kjellr (approved):
Looks good to me! Thank you.

PR #254, opened 2021-04-30
Update Armando

carolinan:


armando/block-template-parts/footer.html
```diff
@@ -1,49 +1,48 @@
...
-		<!-- /wp:paragraph -->
-
-		<!-- wp:buttons {"contentJustification":"center"} -->
-		<div class="wp-block-buttons is-content-justification-center">
-			<!-- wp:button {"borderRadius":3,"backgroundColor":"light-blue","textColor":"dark-blue"} -->
-			<div class="wp-block-button"><a
-					class="wp-block-button__link has-dark-blue-color has-light-blue-background-color has-text-color has-background"
-					style="border-radius:3px">Join our mailing list</a></div>
-			<!-- /wp:button -->
+<!-- wp:group {"align":"full","style":{"color":{"link":"var:preset|color|white"}},"backgroundColor":"dark-blue","textColor":"white"} -->
```
carolinan, line 1: Because styling of template parts is not reflected across the site, I had to move the colors to the group.

armando/style.css
```diff
@@ -152,15 +136,14 @@ we need to adjust them to align them better next to each other. */
...
-}
-
-.blocks-gallery-caption,
-.comment-body {
-	margin-bottom: 2em;
+/* 
+ * Site logo can not be centered on the front without this.
+ * Temporary solution. Needs testing.
+*/
+.wp-block-site-logo.aligncenter {
```
carolinan, line 143: This bug needs to be confirmed, the site logo does not have the same CSS that centers for example the image blocks.

armando/block-templates/page-sidebar-left.html
```diff
@@ -1,46 +1,25 @@
 <!-- wp:template-part {"slug":"header","tagName":"header","className":"site-header is-style-armando-no-bottom-margin"} /-->
-
 <!-- wp:columns {"align":"full","className":"armando-has-sidebar-left is-style-armando-no-bottom-margin"} -->
 <div class="wp-block-columns alignfull armando-has-sidebar-left is-style-armando-no-bottom-margin">
 	<!-- wp:column {"width":"340px"} --><div class="wp-block-column" style="flex-basis:340px">
-	<!-- wp:template-part {"slug":"sidebar-left","tagName":"aside","backgroundColor":"light-blue","className":"armando-sidebar-left"} /-->
+	<!-- wp:template-part {"slug":"sidebar-left","tagName":"aside","backgroundColor":"silver","className":"armando-sidebar-left"} /-->
```
pbking, line 5: I'm not sure I understand why there are two identical sidebar-* template parts.  Perhaps that could be a single template part used by both templates in different ways...

Or maybe you're achieving something I'm not aware of and that's just what you're going for!

armando/classes/class-aboutpage.php
```diff
@@ -0,0 +1,162 @@
...
+						<li><?php esc_html_e( 'Adding images to posts and pages makes your content easier to remember.', 'armando' ); ?></li>
+					</ul>
+				</div>
+			</div>
+		</div>
+		<?php
+	}
+}
+
+new AboutPage();
```
pbking, line 162: What a handy addition!

armando/style.css
pbking: I couldn't confirm that bug.  Removing this style I was able to center the site logo (inside of a group block) just fine without it.

pbking (commented):
Some very nice improvements.  I left only some minor feedback from what I saw; take it for what it is.  Since it's theme-experiments I see no reason not to bring this in as it is, or do with the comments what you will. :)

armando/block-templates/page-sidebar-left.html
carolinan: Originally, there was a template for displaying both a left and right sidebar. I am undecided on what the best way is, perhaps it is best to not use a template part at all, so that different posts/pages can have different sidebars.

armando/experimental-theme.json
```diff
@@ -156,17 +156,27 @@
...
 		"core/post-tags": {
 			"typography": {
-				"fontSize": "var(--wp--preset--font-size--small)"
+				"fontSize": "var(--wp--preset--font-size--extra-small)"
+			}
+		},
+		"core/block-quote": {
+			"border": {
+				"width": "0"
+			},
```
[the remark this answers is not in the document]
carolinan: thank you!

armando/style.css
carolinan: But is it centered also on the font?

PR #256, opened 2021-05-06
Adding bug demonstration

pbking:
This theme demonstrates a bug in Gutenberg where strange things happen.

It seems to have something to do with a block expressing support for a style (or not).  

Or maybe it's coincidental.  

The format (v1 vs v0) doesn't seem to have any bearing.

See [this gutenberg issue](https://github.com/WordPress/gutenberg/issues/31560) for more context.


nosolosw:
Thanks for the test case, hotfix at https://github.com/WordPress/gutenberg/pull/31588

Issue #255, opened 2021-05-04
Divider styled differently in editor vs published site


bbertucc:
In the Site Editor, the divider appears full width:
<img width="1269" alt="Screen Shot 2021-05-04 at 2 27 22 PM" src="https://user-images.githubusercontent.com/46652/117059190-2436ee00-ace5-11eb-8b61-49662e33b066.png">

This is different than the appearance in a published site:
<img width="1288" alt="Screen Shot 2021-05-04 at 2 28 23 PM" src="https://user-images.githubusercontent.com/46652/117059228-3153dd00-ace5-11eb-8107-fe76be75de2d.png">

carolinan:
Hi
I am not able to reproduce this with WordPress 5.7.1, Gutenberg 10.6 and TT1 Blocks 0.4.6.
Instead, nothing happens to the width if I select the wide or full width in the site editor. Which also is incorrect 🤔 

Which version of the theme is used?

carolinan:
The wide and width settings does not work for Twenty Twenty-One either when Gutenberg is active.

carolinan:
I wonder if this is related to de-registering the wide **block** style.

Tested with a few more themes. In Twenty Twenty, the // for the separator is broken, but the widths work.
(I did not check if there is a trac issue for the // style already. Putting it on the to-do list)

bbertucc:
> Hi
> I am not able to reproduce this with WordPress 5.7.1, Gutenberg 10.6 and TT1 Blocks 0.4.6. Windows 10, Chrome.
> Instead, nothing happens to the width if I select the wide or full width in the site editor. Which also is incorrect 🤔
> 
> Which version of the theme is used?

Version: 0.4.5

PR #267, opened 2021-06-01
TT1 Blocks: Make separators 100% wide by default

kjellr:
The thick separators in TT1 Blocks' templates have been broken for a few releases now. They're supposed to stretch full-width, but instead they appear narrow. This doesn't match up with Twenty Twenty-One. 

There's no specific issue for this, though https://github.com/WordPress/theme-experiments/issues/255 seems similar. This PR fixes this via CSS — this behavior is a custom theme decision, so it feels right to do it in CSS. 

Current|Intended
---|---
Screen Shot 2021-06-01 at 8 57 54 AM|Screen Shot 2021-06-01 at 8 56 38 AM

### DIFF
--- a/tt1-blocks/assets/css/blocks.css
+++ b/tt1-blocks/assets/css/blocks.css
@@ -187,6 +187,11 @@ hr.is-style-twentytwentyone-separator-thick,
 	display: none;
 }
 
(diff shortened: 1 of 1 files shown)


danieldudzic (approved):
LGTM

PR #274, opened 2021-07-06
[tt1-blocks] Update deprecated block patterns

ntsekouras:
This PR updates the deprecated block patterns in `tt1-blocks` theme.

The diff is kind of rough and not visible. 

Some of the deprecations were:
1. `Image` blocks with `align` [left, right or center needed a `div` wrapper](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/image/save.js#L68) and the appropriate classes
2. A `Group` block with the extra div `wp-block-group__inner-container`
3. `Column` blocks with a number in `width` like `<!-- wp:column {"width":80}`. The current shape has width as a `string`, so it should be `<!-- wp:column {"width":"80%"}`

## Testing instructions

1. Verify that every pattern's design is not affected and can be inserted with no problems as before
2. Verify that there are no logs produced from deprecated block conversions

In general it should be a good practice to keep block patterns updated and all themes included in core should be updated if needed.

Related: https://github.com/WordPress/gutenberg/issues/33148


ntsekouras:
Putting this to draft state as it has been surfaced here: https://github.com/WordPress/wordpress-develop/pull/1483, that a version check for patterns should be added, besides the version check for the theme itself

PR #275, opened 2021-07-19
emptytheme: use post-template instead of query block

jffng:
`<!-- wp:query-loop -->` is being deprecated, this PR updates it in emptytheme because I was getting a fatal running Gutenberg trunk related to it.


kjellr (approved):
Thanks for the fix!

PR #276, opened 2021-07-21
Remove query-loop block usage from tt1 blocks. Update query markup.

jorgefilipecosta:
This replaces query-loop block usage with post-template block usage (query-loop was deprecated, and is throwing warnings on my tests).
It also updates the query block markup (now it has a `<div class="wp-block-query">`container).
I recommend reviewing the PR without whitespace changes, so it is easier to see what was changed https://github.com/WordPress/theme-experiments/pull/276/files?w=1.


kjellr (approved):
Thanks for the fix, @jorgefilipecosta!

PR #282, opened 2021-08-12
Readme: Add instructions for registering additional theme directory.

iandunn:
AFAIK, this is required to make the themes available. It won't be obvious to everyone, though, so some folks might be confused when they don't see the themes appear in `wp-admin`.


kjellr:
That's weird — I haven't needed do to this to get the themes working before.

iandunn:
huh , it's working in my env w/out this now 🤔

i guess i probably got confused at one point and thought that's why things weren't working, when it was actually something else 🤷🏻

PR #284, opened 2021-08-19
TT1 Blocks: Add border support

kjellr:
This PR adds support for custom borders to TT1 Blocks' theme.json. It also re-creates the "Let's Connect" pattern to use this new feature instead of the old custom block style. 

None of the other border patterns can be replaced yet, since the border settings required are only available for the Group block presently. 

Screen Shot 2021-08-19 at 10 54 05 AM


tt1-blocks/theme.json
```diff
@@ -11,6 +11,12 @@
 		}
 	],
 	"settings": {
+		"border": {
+			"customColor": true,
+			"customRadius": true,
+			"customStyle": true,
+			"customWidth": true
+		},
```
pbking, line 19: Looks like border customizations are [already enabled](https://github.com/WordPress/theme-experiments/blob/203ee43174c61968b0d71111b38444aa199fbc7d/tt1-blocks/theme.json#L194) for the theme.

kjellr: 😄 Thanks. I've removed that.

kjellr:
Since it turned out that custom borders were already active in the theme, I've reframed this PR as just a refresh of that single block pattern instead.

pbking:
It was a little odd that the padding was in the columns and not in the group.  



When I first dropped it in I tried to tweak that at the top level container and found I couldn't.

PR #285, opened 2021-08-19
TT1 Blocks: Enqueue stylesheets instead of using @import

### LINKED ISSUE #198: TT1 Blocks: Don't use @import inside a css file
https://github.com/WordPress/theme-experiments/blob/7da54cad8dfd0a5ea1726fc3745e94a9802d3892/tt1-blocks/style.css#L24

It should never be used `@import` inside a CSS file.

This creates two problems, one is a bad practice for performance reasons and another is that a junior developer may think it is a good practice and start write CSS like that.

kjellr:
Fixes #198 

This PR enqueues the theme's additional stylesheets (`blocks.css` and `style-shared.css`) rather than importing them. This is [generally a better practice](https://wordpress.stackexchange.com/questions/167064/why-shouldnt-import-be-used-to-import-a-parent-theme-into-a-child-theme).


tt1-blocks/functions.php
```diff
@@ -69,6 +69,8 @@ function tt1_blocks_setup() {
  */
 function tt1_blocks_scripts() {
```
Ismail-elkorchi, line 70: Since this function will enqueue the css files, it's name must be changed to reflect the actual situation. `tt1_blocks_scripts_and_styles()` seems like a good fit.

kjellr: I've updated this to `tt1_blocks_enqueue_styles` which I think is more descriptive (this function enqueues just stylesheets, no scripts). 

For what it's worth, the original `_blocks_scripts` function name was brought over from Twenty Twenty-One, and in that case, the function does enqueue both scripts and styles.

carolinan (approved):

PR #293, opened 2021-11-23
TT1-Blocks: Update Query block in index template to use displayLayout instead of layout

andrewserong:
On a fresh install of Gutenberg, running in `wp-env`, I noticed that the current `index.html` template throws an error, due to the `layout` attribute being set to an invalid value (`list`):

This then throws an error in `useAvailableAlignments` [at this line](https://github.com/wordpress/gutenberg/blob/32ecd97e988174b6401289c44200135c224ca104/packages/block-editor/src/components/block-alignment-control/use-available-alignments.js#L34)

image

This is because the `layout` attribute in this markup should be `displayLayout`, as `layout` is reserved for the Layout block support. The switch to `displayLayout` was introduced in https://github.com/WordPress/gutenberg/pull/31833 (back in May).

It seems that this error occurs because for the current markup the `layout` attribute is being used directly in the Query block [here](https://github.com/wordpress/gutenberg/blob/4d3fb72d301f26bcf13f3117cec7e5abe60f399e/packages/block-library/src/query/edit/index.js#L47) — and because the current markup includes the container `div`, it never gets migrated via a deprecation (which [usually handles the migration](https://github.com/wordpress/gutenberg/blob/bf405a3f1280cade6b9f480c973261b987d5fd53/packages/block-library/src/query/deprecated.js#L49) from `layout` to `displayLayout`).

Just to double-check before we do any further work in Gutenberg, @jorgefilipecosta it looks like this markup was last updated in https://github.com/WordPress/theme-experiments/pull/276 — do you remember if the updated markup was copy + pasted or hand-edited? Just trying to work out whether the markup here had a typo, or if we need additional deprecations in Gutenberg to handle this case?

### Steps to test

With Gutenberg activated, go to create a new page and copy and paste the existing markup for the `wp:query` block [from this file](https://github.com/WordPress/theme-experiments/blob/master/tt1-blocks/block-templates/index.html) into the code editor view. You should see the error message in the screenshot above when you switch back to the visual editor.

Edit the markup in the code editor view and copy paste the change in this PR over the existing markup. When you switch back to the visual editor view, there should no longer be an error. Or, alternately, spin up a new WP instance and use this PR to load the TT1-Blocks theme. When you open the site editor, there should be no errors in the Query block.

### DIFF
--- a/tt1-blocks/block-templates/index.html
+++ b/tt1-blocks/block-templates/index.html
@@ -2,7 +2,7 @@
 
 <!-- wp:group {"tagName":"main"} -->
 <main class="wp-block-group">
-	<!-- wp:query {"queryId":1,"query":{"perPage":"10","pages":"100","offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":"","inherit":true},"layout":{"type":"list"}} -->
+	<!-- wp:query {"queryId":1,"query":{"perPage":"10","pages":"100","offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":"","inherit":true},"displayLayout":{"type":"list"}} -->
 	<div class="wp-block-query">
 		<!-- wp:post-template -->
 


andrewserong:
CC: @kjellr, too. I don't have access to add reviewers or labels to this repo 🙂

carolinan:
Since the theme has not been updated since June I would expect there to be a large amount of compatibility issues.
I don't think anyone frankly has time to update the theme until after we are all rested after 5.9.

Since Twenty Twenty-Two will be the new theme that is tested with Gutenberg, can we priorities making that change instead?

carolinan:
While I can merge it, I don't know how to tag it as a new release, and I can't release it to .org.

andrewserong:
Thanks for merging @carolinan! And good to know that it's expected there'll be issues with TT1-Blocks until 11.9 comes out. I'll keep that in mind and focus more of my testing on TwentyTwentyTwo instead. Cheers!

andrewserong:
As an aside, I just took a quick look at the TT2 templates and it looks like they don't have the same issue (the `layout` attribute in those templates appears to be set correctly) 👍

carolinan:
With change, I meant, make sure that Gutenberg uses TT2 not tt1-blocks :) https://github.com/WordPress/twentytwentytwo/issues/133

andrewserong:
Perfect, thanks for confirming @jorgefilipecosta, just wanted to make sure I didn't need to investigate the block itself further in the Gutenberg repo, and that confirms it. Cheers! 🙇

PR #294, opened 2021-12-15
Refresh emptytheme

kjellr:
This does a few things to bring `emptytheme` more in sync with latest Gutenberg developments: 

- It changes the `theme.json` version to 2.
- It uses `"appearanceTools": true` instead of opting into a bunch of different features.
- It changes `block-templates` and `block-template-parts` folders to `templates` and `parts`.
- It removes some `theme_supports` that now are automatically added to block themes.


scruffian:
Thanks

PR #295, opened 2021-12-16
Emptytheme: Stop clearing presets

kjellr:
Emptytheme contains the following code in its theme.json, which is a helpful starting place for theme authors to populate:

```
		"color": {
			"gradients": [],
			"palette": []
		},
		"typography": {
			"fontFamilies": [],
			"fontSizes": []
		},
```

... but the downside is that this clears out those presets, and makes their associated tools unavailable (as noticed by @jasmussen [here](https://github.com/WordPress/gutenberg/pull/37446#issuecomment-995608534)). 

I think it makes sense to remove them.

### DIFF
--- a/emptytheme/theme.json
+++ b/emptytheme/theme.json
@@ -2,14 +2,6 @@
 	"version": 2,
 	"settings": {
 		"appearanceTools": true,
-		"color": {
-			"gradients": [],
-			"palette": []
-		},
-		"typography": {
-			"fontFamilies": [],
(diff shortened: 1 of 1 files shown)


MaggieCabrera (approved):
I agree, I thought about it but never got to make a PR, LGTM!

jasmussen:
Nice one. I was about to ask for the default color palette to also work, but you're on top of it. Sweet!

PR #296, opened 2021-12-17
Maintain changelog and readme file

kafleg:
This version will fix the issues of the query-loop to post-template.

### DIFF
--- a/tt1-blocks/readme.txt
+++ b/tt1-blocks/readme.txt
@@ -1,9 +1,9 @@
 === TT1 Blocks ===
 Contributors: wordpressdotorg
 Requires at least: 5.6
-Tested up to: 5.7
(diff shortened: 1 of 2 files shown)


tt1-blocks/readme.txt
```diff
@@ -1,9 +1,9 @@
 === TT1 Blocks ===
 Contributors: wordpressdotorg
 Requires at least: 5.6
-Tested up to: 5.7
+Tested up to: 5.8
```
carolinan, line 4: Should this be 5.9?

kafleg: 5.9 is still beta. I think we need to update it when 5.9 is available.

kjellr (approved):
Thanks. We can push another update after 5.9.

PR #297, opened 2021-12-17
Change directory name

kafleg:
block-template-parts and block-template directory name changed


carolinan:
Will this work for existing installations on WordPress versions lower than 5.9?
Hmm I suppose it depends if the user has updated Gutenberg.

kafleg:
> Will this work for existing installations on WordPress versions lower than 5.9? Hmm I suppose it depends if the user has updated Gutenberg.

Tested with WP 5.8.2 and Gutenberg Version 12.1.0 
It is working fine.

kjellr:
Since these directory names will still work even in the new versions of Gutenberg, I'm going to hold off from deploying this as part of the 0.4.8 release. We can bundle this in the next one and give folks a little more time to update.

PR #299, opened 2021-12-17
Fix TT1 Blocks layout issues

### LINKED ISSUE #298: TT1 Blocks: Visual issues
There are a number of glaring visual issues in TT1-Blocks at the moment: 

- [ ] Content bumps up against the edges of the screen.
- [ ] The wide-aligned header, footer, and post titles are all full-width.
- [ ] The default navigation is not right-aligned.

These are likely due to old markup. We bundled up a new version in #296, but after testing the zip before merge, I'm not comfortable pushing a "new" version that looks this broken still. Let's instead fix these issues before we sync up with WordPress.org. 

## Screenshot

Screen Shot 2021-12-17 at 7 46 09 AM

kjellr:
Fixes #298. 

- Template parts no longer inherit the default layout, so I've moved that into the  templates themselves in the form of an extra group wrapper.
- I've updated the navigation block markup to be in sync with latest versions. 
- I've ported over the left/right padding solution from Twenty Twenty-Two. I don't love this being duplicated in multiple themes, but it's what we've got at the moment. 

cc @kafleg 

Before|After
---|---
Screen Shot 2021-12-17 at 11 03 50 AM|Screen Shot 2021-12-17 at 11 05 12 AM

### DIFF
--- a/tt1-blocks/assets/css/style-shared.css
+++ b/tt1-blocks/assets/css/style-shared.css
@@ -30,3 +30,42 @@ a:hover {
 .site-header h1.wp-block-site-title a:not(:hover):not(:focus):not(:active) {
 	text-decoration: underline;
 }
+
+/*
+ * Alignment styles.
+ * These rules are temporary, and should not be relied on or 
+ * modified too heavily by themes or plugins that build on 
+ * Twenty Twenty-Two. These are meant to be a precursor to 
+ * a global solution provided by the Block Editor. 
+ * 
+ * Relevant issues:
+ * https://github.com/WordPress/gutenberg/issues/35607
+ * https://github.com/WordPress/gutenberg/issues/35884
+ */
+
+.wp-site-blocks,
+body > .is-root-container,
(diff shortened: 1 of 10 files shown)


MaggieCabrera (approved):
This LGTM, I wonder when will this theme be ready for a v1? if we make it ready for 5.9 maybe?

kjellr:
Thanks! I'm not sure about a v1! Probably needs a little more attention and cleanup before then.

kjellr:
@MaggieCabrera mind giving it one more look after https://github.com/WordPress/theme-experiments/pull/299/commits/0248eba4cf10fc83dc83af315e4a15b945eb69f8?

kafleg:
This PR looks good to me. Thank you @kjellr

MaggieCabrera:
> @MaggieCabrera mind giving it one more look after [0248eba](https://github.com/WordPress/theme-experiments/commit/0248eba4cf10fc83dc83af315e4a15b945eb69f8)?

yes, that looks much better, thanks! 🚢

PR #300, opened 2021-12-17
TT1 Blocks: fixed query pagination markup

MaggieCabrera:
The query pagination no longer needs a div wrapper, this PR removes it and adds space-between justification to the content.

Before:

Screenshot 2021-12-17 at 17 23 51

After:

Screenshot 2021-12-17 at 17 23 40


kjellr (approved):
Thanks!

PR #302, opened 2022-02-09
Demonstrate an issue with the query pagination block

scruffian:
This adds a query pagination block to emptytheme, inside a template. Notice that when you click "next page" the content on the page stays the same.


scruffian:
Closing this old PR.

PR #305, opened 2022-07-06
TT1 blocks: implement new comments block

MaggieCabrera:
This PR implements the new comments block in TT1 blocks

Before:
Screenshot 2022-07-06 at 10 27 37

After:
Screenshot 2022-07-06 at 10 28 11

In other themes I've made this into a hidden pattern to avoid code repetition, I'm not sure if we should do the same for TT1b. I'm open to suggestions.


carolinan:
All of this could be done in a follow up PR, or, if you want to continue working on it:

- Yes, the comments area could be a pattern or a template part.
- If we are increasing the minimum required theme version, which I hope, then the template should be moved to the `templates` folder
- The page needs some spacing between the content and the comments.


----
Comments area of twenty twenty-one: 
https://wp-themes.com/twentytwentyone/?p=1#comments

Tt1 blocks, before PR:
https://wp-themes.com/tt1-blocks/?p=1#comments

MaggieCabrera:
Yeah I can add those changes to this PR

carolinan:
Aside: Ha, WP does not understand what to do when there is both a block-templates and a templates folder

Issue #193, opened 2021-02-03
TT1 Blocks: Changing Image height/width settings unresponsive


annezazu:
I tested in both TT1 and Twenty Twenty-one and found the issue was only in TT1. Essentially, manually changing the width/height up or down seems to cause some odd behavior. Here's a quick video:

https://user-images.githubusercontent.com/26996883/106691606-05637700-6591-11eb-8351-23c1c8350ca5.mov

Apologies for the odd title :)  This is using WordPress 5.6 and Gutenberg Version 9.9.0.20210201 ([a nightly build](https://gutenbergtimes.com/need-a-zip-from-master/#nightly) from the Gutenberg Times).

carolinan:
I have been able to reproduce this with TT1 Blocks, WordPress 5.7.1 and Gutenberg 10.6.

It is not reproduceable with emptytheme, Q, Armando or Quadrat.

youknowriad:
Sounds more like a Gutenberg issue can we move this one there maybe?

carolinan:
It seems to be related to this theme CSS 

```
.wp-block-image img {
	height: auto;
}
```
https://github.com/WordPress/theme-experiments/blob/master/tt1-blocks/assets/css/blocks.css#L53

PR #307, opened 2022-07-06
TT1-Blocks update

carolinan:
- Moved templates and parts to the new folders
- Moved patterns to the new patterns folder
- Updated theme.json to version 2
- Removed some CSS that is no longer needed.
- Removed theme_support that is no longer needed, since it is added automatically.

### DIFF
--- a/tt1-blocks/readme.txt
+++ b/tt1-blocks/readme.txt
@@ -3,7 +3,6 @@ Contributors: wordpressdotorg
 Requires at least: 6.0
 Tested up to: 6.0
 Requires PHP: 5.6
-Stable tag: 0.4.9
 License: GPLv2 or later
 License URI: http://www.gnu.org/licenses/gpl-2.0.html
 


carolinan:
> I think this one closes #193

Tested, confirmed (chrome, windows 10)

carolinan:
Testing the patterns, it looks like `is-style-twentytwentyone-border` does not work in the Site Editor.

tt1-blocks/style.css
```diff
@@ -4,10 +4,10 @@ Theme URI: https://github.com/WordPress/theme-experiments/tree/master/tt1-blocks
 Author: the WordPress team
 Author URI: https://wordpress.org/
 Description: TT1 Blocks is an experimental block-based version of the Twenty Twenty-One theme. It is built to leverage the full-site editing functionality that is being built in the Gutenberg plugin. This theme is not meant for use on a production site.
-Requires at least: 5.6
-Tested up to: 5.8
+Requires at least: 6.0
+Tested up to: 6.0
 Requires PHP: 5.6
-Version: 0.4.8
+Version: 0.4.9
```
MaggieCabrera, line 10: do you think these changes justify the bump to the 1.0.0 version? I don't think we should consider tt1blocks experimental anymore if tt2 isn't either

carolinan: Yeah we could do that,

tt1-blocks/readme.txt
```diff
@@ -1,9 +1,9 @@
 === TT1 Blocks ===
 Contributors: wordpressdotorg
-Requires at least: 5.6
-Tested up to: 5.8
+Requires at least: 6.0
+Tested up to: 6.0
 Requires PHP: 5.6
-Stable tag: 0.4.8
+Stable tag: 0.4.9
```
carolinan, line 6:
```suggestion
```

Theme's don't user "stable tag", only plugins do.

tt1-blocks/patterns/large-text.php
```diff
@@ -0,0 +1,17 @@
+<?php
+/**
+ * Title: Large text
+ * Slug: tt1-blocks/large-text
+ * Categories: tt1-blocks
+ * Block Types: core/heading
+ * ViewportWidth: 1440
```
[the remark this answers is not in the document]
carolinan: I don't know if it is a true bug or, if there is a limit to how many patterns are loaded from the directory, or something; but it does not work as I expected, if Gutenberg is active:
https://github.com/WordPress/gutenberg/issues/41959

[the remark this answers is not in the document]
MaggieCabrera: mmh, I guess we could raise this in core-editor chat? I don't know who works on this. Maybe we could follow up on a future PR.

MaggieCabrera:
Looks like we lost the separator style on index? I don't know if this is still a WIP, let me know for a full review

Screenshot 2022-07-06 at 14 46 30

